PR #9, opened 2019-07-19
Add Google Analytics Support

devbali:
Using the supported option given by the Sphinx Read the Docs theme, adding Google Analytics support lets us have better data about who is coming to read Zeek Documentation.


jsiwek:
Do you also need this to be added to the "stable" version of the docs for v2.6.2 ?  I'll need to commit it to a different branch and make a release tag for that (not hard, making sure what we need).

rsmmr:
Yeah, that would be good.

PR #15, opened 2019-10-03
Require C++17

Neverlord:


Neverlord:
I've removed the `alternatives` step, since it's no longer necessary with https://github.com/zeek/zeek/pull/645.

PR #47, opened 2020-10-17
documenting (ZAM_)vector semantics

vpax:
Related to https://github.com/zeek/zeek/pull/1231

### DIFF
--- a/script-reference/types.rst
+++ b/script-reference/types.rst
@@ -706,7 +706,7 @@ A range of elements can be *replaced* by assigning to a vector slice:
     v[0:2] = vector("five", "six", "seven");
 
 The size of a vector (this is one greater than the highest index value, and
-is normally equal to the number of elements in the vector) can be obtained
+is equal to the number of elements in the vector) can be obtained
 by placing the vector identifier between vertical pipe characters:
 
 .. code-block:: zeek
@@ -730,7 +730,8 @@ this expression is equivalent to:
 The ``in`` operator can be used to check if a value has been assigned at a
 specified index value in the vector.  For example, if a vector has size 4,
 then the expression ``3 in v`` would yield true and ``4 in v`` would yield
-false.
+false.  The result of accessing a vector index that is not "in" the vector
+is undefined.
 
 Vectors of integral types (``int`` or ``count``) support the pre-increment
 (``++``) and pre-decrement operators (``--``), which will increment or
@@ -750,6 +751,11 @@ each element of the operand vectors.
 
 Vectors of type ``count`` can also be operands for the bitwise and/or/xor
 operators, ``&``, ``|`` and ``^``.
+
+Vectors of type ``any`` can hold values of any Zeek type, but the values
+must be *homogeneous*.  If you want to intermingle values of different types,
+you can instead use ``table[count] of any``, which does not require
+homogeneous values.
 
 See the :zeek:keyword:`for` statement for info on how to iterate over
 the elements in a vector.
--- a/script-reference/types.rst
+++ b/script-reference/types.rst
@@ -752,11 +752,6 @@ each element of the operand vectors.
 Vectors of type ``count`` can also be operands for the bitwise and/or/xor
 operators, ``&``, ``|`` and ``^``.
 
-Vectors of type ``any`` can hold values of any Zeek type, but the values
-must be *homogeneous*.  If you want to intermingle values of different types,
-you can instead use ``table[count] of any``, which does not require
-homogeneous values.
-
 See the :zeek:keyword:`for` statement for info on how to iterate over
 the elements in a vector.
 


vpax:
Somehow my original comment, "companion to topic/vern/brovalunion-change", got lost

jsiwek:
@vpax is this PR still planned to get incorporated in some form later or ok to close for now?

vpax:
@jsiwek I think this documentation change will remain apt once we integrate the `ZVal` changes (new representations for records and vectors) that I'm going to work with @timwoj on fairly soon.

jsiwek:
@vpax and/or @timwoj is the status of these doc changes any different/applicable now ?

vpax:
@jsiwek The changes (using both commits, one of which undoes a change in the other) are apt to now integrate.  Basically, this is codifying that the undocumented feature of vector "holes" is undefined behavior.

jsiwek:
@vpax ok, then trying to understand a bit more about it, and whether something is need in NEWS for any change in functionality:

* The doc changes here don't mention anything about creation of "holes", just the accessing of non-existent indices, so would it help to distinguish them as different situations?  For example, if creating a "hole" is the thing that's actually undefined, helps to say that because the creation of a hole may actually lead to `in` operator reporting the index exists, leaving one to wonder what they actually did wrong.

* Is the behavior of accessing non-existent index different than it was before?  Practically, that still causes runtime error/exception (unwind out of current event handler)?  I can't find if that behavior is documented (also don't see it mentioned for `table`/`set` either), so I guess it's fine to now call out as undefined, but wondering about whether there's reservations about actually defining it?  If it's important for performance/optimization stuff later, maybe fine, but it's also common mistake, so often helpful to understand what kind of behavior to expect when debugging and troubleshooting.

vpax:
@jsiwek Good questions, and they led me to realize that these documentation tweaks aren't really on point.

Here's what I'm aiming to capture.  The old semantics were that assigning to an element more than 1 beyond the end of a vector would populate the stepped-over values with "holes", meaning missing values.  These semantics are not documented explicitly but are implicitly present in the discussion that `The size of a vector ... is _normally_ equal to the number of elements in the vector` (emphasis added), and, somewhat, in `The "in" operator can be used to check if a value has been assigned at a specified index value in the vector`.  (Here I say "somewhat" because for the example it gives, the `false` instance is beyond the end of the vector, not a hole inside the vector.)

In the new semantics, the treatment of holes becomes undefined.  A key part of this, not correctly framed in the current documentation tweaks, is that the `in` operator does _not_ any more track whether a given vector element has been assigned too. It now simply equates to "is this index between zero and the size-of-the-vector-minus-one, inclusive".

This difference in semantics is for performance reasons. For the old vector implementation, holes came "for free".  For the new implementation, they don't (well, there are circumstances where they do, but they don't always apply).  We could modify `VectorVal` to have a parallel bit array that tracks whether elements have been set (`RecordVal` does this), but it seems a pity to spend the storage and computation on that to support a feature that has never been explicitly documented, and has only marginal utility.  One can always get the same functionality using a `table[count] of X` instead.

jsiwek:
@vpax thanks, I get it now and summarized in release notes like this: https://github.com/zeek/zeek/commit/361c99cc26cef1391c97b2cc39f3b4acb365b814.

Do you plan to make tweaks to the docs themselves (revise this PR) ?

0xxon:
I probably lost my chance to push against this change with not noticing this in #1231. But just for the record - I think that the change of not allowing the test for holes will lead to a lot of hard to find bugs. It also might be expected behavior in user scripts - which worked for years and years. And - if I am understanding everything correctly - we are breaking it without deprecating it, and without warning - which is against our deprecation policy.

I have no idea if this makes any difference - but I know at least one place in the scripts that we ship with Zeek that will regularly produce vectors with holes (which then in this case, as far as I remember, only are sent on to the logging framework).

I also use the fact that you can check for holes in scripts that are running at sites. Suddenly declaring this previously supported (albeit not documented - but there has historically been a lot that is underdocumented) seems... at least not nice.

So - short form - I think that we should _at least_ detect this behavior and warn people having this in their scripts for a full release cycle. Optimally - I like the old behavior...

0xxon:
Also - this is an impactful enough change that it seems like it merits a discussion that is visible to a wider audience - like the zeek-dev mailing list.

0xxon:
Ok, now that it has been two hours or so - I thought about this a bit more. And I am even more on the side of strongly vetoing this change.

One of the typical cases that I use vectors for is construct a vector based on something that happens in a protocol in sequence, and then directly send that on to a log file.

In the cases where I use it, if someone violates the protocol, there can be holes in the vector. These holes have semantic meaning (something happened here - it did not fit the spec, so we cannot log it as typically, but there was a message at position X that does not fit the scheme). Currently you can stuff this in a vector with a hole, send it on to the logging framework, and then, in the log, you can nicely recognize this.

The proposal of using a table[count] does not work for this case at all - since you cannot log it. Neither can you use a set (you loose ordering). So - if this change goes through, I will first have to create a table[count] in scriptland, and then for logging it, loop over it and create a string-representation from it, and then send that to the logging framework as a string.  Or - insert dummy elements (which is not great in other ways), or have a separate entry in the log that tracks these holes.

I am sure there are other valid use-cases for this. So...

* I would strongly prefer this change not to be made
* if do decide that we want to make this change, we have to follow our [deprecation policy](https://github.com/zeek/zeek/wiki/Release-Cadence#backwards-compatibility) and deprecate this in Zeek 5, and then can remove it in Zeek 6 - following our deprecation policy. In my opinion we should not change such a basic behavior that has worked forever without any pre-warning.

vpax:
@0xxon I'd like to better understand your use case.

1. What sorts of values do you store in the vector?  (One thing I'm wondering is whether this is a `vector of any`.)
2. When you have an absent value, do you know its index?  If so, could you store a sentinel value to denote "this is missing", rather than not assigning anything to that location?
3. What would you think about introducing an attribute like `&sparse` which would mean "this vector should support holes"?

0xxon:
>     1. What sorts of values do you store in the vector?  (One thing I'm wondering is whether this is a `vector of any`.)

In the cases that I can recall doing - numbers and strings. For strings, putting in placeholder values is actually relatively easy. For numbers it gets a tad complicated - I think some of them are 64 bit uints, so whatever you choose there is some chance for a collision.

>     2. When you have an absent value, do you know its index?  If so, could you store a sentinel value to denote "this is missing", rather than not assigning anything to that location?

See above. So - yes, kind of, but... it is not really nice.

>     3. What would you think about introducing an attribute like `&sparse` which would mean "this vector should support holes"?

I would prefer that to not having the support anymore. But - I honestly still would prefer the support to stay in. And I also still think that this needs a full deprecation cycle. While this was not documented as such - this is not exactly a super unknown feature. So I can see this being used in quite a few pieces of code.

To make things even more complex - at least for me this is in codepaths that are a bit unusual - for me it only happens in slightly weird protocol flows - so if this change was introduced without me noticing I would probably have spent quite a while debugging it.

vpax:
I hear you re the deprecation concerns.

FWIW, the new actual semantics are that holes continue to work for all but numeric types (where numeric means the underlying representation is numeric, so includes things like `port`, `enum`, `interval`, `time`).  For numeric types, access to the hole won't fault, it'll instead yield a zero (of the numeric type).  It would be clunky to make this the defined semantics, though.

Also FWIW, a `vector of any` will provide the current semantics, and I could see documenting holes as explicitly allowed for those.  I imagine however that you wouldn't want to lose static type-checking by switching to using those.

I advocate for going the attribute route (or perhaps the `vector of any` route) because it seems quite a pity to me to slow down all vector operations in support of a not commonly used feature.

jsiwek:
> I advocate for going the attribute route (or perhaps the `vector of any` route)

A couple alternate ideas:

* Generalize a way to express nullability/optionality for all types/variables, not just via `&optional` record fields.  Maybe that could actually re-use `&optional`, but trying to associate attributes within compound types can be ambiguous, so would rather suggest a new, unambiguous syntax.  Then, this can get you something like a `vector of optional<count>`, which explicitly expresses the nullability intent.

* Allow logging of `vector of SingleFieldRecord`, which gives the potential for making that single-field have the `&optional` attribute.

> it seems quite a pity to me to slow down all vector operations in support of a not commonly used feature.

Is the slowness of allowing this behavior in comparison to just previous implementation or in comparison to new implementations "full potential" ?  And any estimate of how much slower?

If not much slower than previously, having temporary slowness for duration of the deprecation period may be acceptable?

Side note: noticed there is a btest relying on vector holes that happens to still work as intended because it uses strings: https://github.com/zeek/zeek/blob/1bbb7fc4bd5c997d72dd6085bad471fe2b04dfc4/testing/btest/scripts/base/frameworks/logging/vec.zeek

vpax:
`vector of optional T` seems fine to me, though I'd rather we fix attributes to not have the current ambiguity problems, since that has broader utility.  That said, in what other situations would it make sense to use an `optional` declaration? I'm not offhand picturing it as having generality, but maybe I'm missing some natural idioms.

The `vector of SingleFieldRecord` approach seems okay to me only if the usage is indeed quite rare, because it's somewhat clunky.

The slowness is in comparison to the new implementation's "full potential".  It would make vector element assignments/accesses require two underlying `std::vector` assignments/accesses rather than one. The additional one would be a tad more involved, too, if it's done using the `std::vector<bool>` specialization in order to use less memory. For assignments, there's some additional code that would stay the same, so the net cost would be < 2x.  For accesses, there's not really any additional code, so it would be around 2x (a bit higher for the case where the element is missing, but presumably that's much rarer than it being present). There would be a memory cost, too.  OTOH, from the profiling I did for ZAM, for the scripts I assessed, vector operations weren't common, so the end-to-end cost is likely quite low, and wouldn't be a problem to abide during a deprecation period.

jsiwek:
> That said, in what other situations would it make sense to use an `optional` declaration?

Thinking it's any situation in which an `&optional` record field is used currently, but happens to be data that's never closely associated with anything else, so you end up with a single-optional-field-record (that's clunky).  I'd struggle to give exact places that's used or the few times I may have suggested that pattern to others (one I do remember is `Broker::Data`), so I guess it maybe is uncommon.  Still interesting how function parameters are currently so closely associated w/ `RecordType` that it's possible to declare an `&optional` function/event parameter, but can't currently make use of it (no way to pass in a "nil" and no way to check existence).

Another place it could be useful is for exceptional/error return values from functions.  Can currently look through existing functions (e.g. recall this pattern mostly in BIFs) to find many that may not do a proper job of reporting errors to callers besides having the side-effect a `reporter.log` message, or else rely on a sentinel value to indicate that an error happened (e.g. `strftime` returning `"<strftime error>"`).  An optional/null return value could be more convenient for those.

0xxon:
> I hear you re the deprecation concerns.

Honestly - I think this alone tops everything. We have a deprecation policy - and for a feature that has been working for at least a decade (and probably much longer) - we should definitely stick to this. I get that it is annoying that this makes the implementation slower and more complex - but I don't see that overriding the fact that this would be a quite big change in semantics that we are making without any pre-warning and without any good way of people being able to detect it.

How about turning this around - and having an (potentially temporary) way of opting in to have vectors where one guarantees that they won't have a hole?

rsmmr:
> We have a deprecation policy - and for a feature that has been working for at least a decade (and probably much longer) - we should definitely stick to this.

I'm with Johanna on this. I admit that I didn't think too much about this change at first, but realizing  that there are scripts out there that are relying on our current semantics (including one of our own tests), and that it's quite difficult to even spot the places affected (with the risk of scripts breaking subtly with rare traffic patterns), I agree that we need to be more careful here. And the fact that the concrete semantics depend on the type being stored, doesn't really make it better IMO.

Opt-in to the new semantics would be one approach, but I'm wondering if simply retaining the old way is really that much of a problem. Vern, you wrote earlier:

> We could modify VectorVal to have a parallel bit array that tracks whether elements have been set

What if we instead stored a `std::vector<std::optional<ZVal>>`? That'd be easy to work with, and my guess is that performance impact would be negligible because (1) vector operations are just not that common/massive overall that they would dominate performance, and (2) it's at least a  more cache friendly approach to storing that additional bit of information.

vpax:
> What if we instead stored a `std::vector<std::optional<ZVal>>`?

Nice thought!  I like your observation about cache-friendly-ness.  Let me take a stab at implementing that.

jsiwek:
No change seems needed after https://github.com/zeek/zeek/pull/1512 gets merged so closing this (or if there are, just make a new PR).

PR #55, opened 2020-12-17
Debian 10: error: Could not find prerequisite package 'PythonDev'

ChrisSwanson:
Previously receiving error mentioned above since libpython3-dev package contents appeared to be missing in Debian 10 (buster).

### DIFF
--- a/install/install.rst
+++ b/install/install.rst
@@ -67,7 +67,7 @@ To install the required dependencies, you can use:
 
   .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


jsiwek:
The `python3-dev` package, already included in the documented command, is supposed to automatically depend on `libpython3-dev`: https://packages.debian.org/buster/python3-dev

So could that have just been a transient error in Debian packaging, package manager, or your local system?

ChrisSwanson:
Yeah, looked a little further into this on my end, after you mentioned this, looks like this error was more localized, and for whatever reason the initial installation of libpython3-dev didn't take upon initial compile tries.  Thanks for the heads up!

PR #58, opened 2021-03-23
Illustrate required CMAKE_PREFIX_PATH on MacOS ./configure

bsolomon1124:
This is virtually _required_ on MacOS, since it ships with Bison 2.3, a version that `./configure` will fail with unless CMake is made aware of where the Homebrew install of Bison is. Force-linking is _not_ the solution :)

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -107,6 +107,16 @@ To install the required dependencies, you can use:
   ``swig``, ``swig-python``, ``openssl``, and ``bison`` packages provide
   the required dependencies.
 
+  Note that MacOS may ship with an outdated version of Bison that may
+  take precedence over the newer version of Bison installed with
+  a package manager. To use the (required) newer version, set
+  ``CMAKE_PREFIX_PATH``; for example:
+
+  .. code-block:: console
+
+     brew install bison cmake swig openssl
+     CMAKE_PREFIX_PATH="$(brew --prefix bison)" ./configure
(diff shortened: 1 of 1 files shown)


jsiwek:
Thanks, but the following CMake logic is supposed to help automatically figure this out such that a simple `./configure` generally works:

https://github.com/zeek/cmake/blob/74259745dea5ee4889d1ac1f4ebde4e2c59c329a/MacDependencyPaths.cmake#L25

So would like to understand where that went wrong.  Can you please make a separate issue regarding the default `./configure` failing to find Homebrew's bison for you, giving the output of your `brew --prefix bison` as a starting point ?  Or make a PR to improve the logic linked above if you spot what's wrong ?

jsiwek:
And just to guess randomly: the paths containing `${BREW_PREFIX}` in the CMake logic may need to be within quotes to work properly with a `brew --prefix` that contains spaces.

bsolomon1124:
Welp, I went to make an MVCE to file a separate issue, and could not reproduce my own problem :(

After ensuring the latest and greatest `bison`, `swig`, and `cmake` were installed via `brew` and restarting my shell, I did not need `CMAKE_PREFIX_PATH` to configure successfully.

Thank you for taking a look.

Btw, my `brew --prefix bison` is `/usr/local/opt/bison`.

PR #86, opened 2021-07-27
GH-72: Add section about script usage errors

### LINKED ISSUE #72: Cover the new attributes and new functionality relating to use defs
@vpax could you help me out here? Is the work of zeek/zeek#1394 at a point where you'd prefer to see it covered in the docs? I tagged it because it seemed that way, but I've not been following this closely. Thx.

timwoj:
Fixes #72

### DIFF
--- a/scripting/basics.rst
+++ b/scripting/basics.rst
@@ -1,9 +1,9 @@
 
 .. _writing-scripts:
 
-=========================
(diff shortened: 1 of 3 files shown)


taosecurity (approved):
Good to go.

PR #90, opened 2021-08-25
Edit pass over documentation of callable types

ckreibich:
I figured out last week that functions using `return when` pass their arguments by value, not reference. Trying to document this, I ended up editing callables more broadly. Feedback welcome:

- Introduce "asynchronous function" as the name of functions that use "return
  when". It looks like they didn't really have an official term yet, and
  this one relates it to the broadly similar concepts e.g. in JavaScript and
  Python.

- Expand the table of callable-types features to state that asynchronous
  functions pass arguments by value.

- Structure the documentation of functions from one big body of text into three
  sections: anonymous functions and their closures, default values, and
  asynchronous functions.

- A bit of optical polish and line breaks of the callable-types table and
  descriptions, with additional linking and minor corrections.

@rsmmr, @vpax, @anthonykasza — fyi!

### DIFF
--- a/script-reference/statements.rst
+++ b/script-reference/statements.rst
@@ -40,7 +40,7 @@ all loaded Zeek scripts.
   * - :zeek:keyword:`redef`
     - Redefine a global value or extend a user-defined type
 
-  * - `function/event/hook`_
+  * - `Callables`_: :zeek:type:`function`, :zeek:type:`event`, :zeek:type:`hook`
     - Declare a function, event handler, or hook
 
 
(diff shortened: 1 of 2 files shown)


rsmmr:
Thanks for documenting this, I like the name "asynchronous function", which will eventually also generalize if we ever extend this concept further (e.g., to no longer require `when`). 

One clarification though: I don't think these "asynchronous functions" really pass their arguments by value. What happens is that the body of the `when` captures a closure of the surrounding scope, so arguments are passed by value *into that body*. But if the function has logic working on any function parameters before the `return when`, then those have the normal semantics. At least that's how I remember this (and adding a couple of tests to confirm would probably be good. :-) ).

ckreibich:
Ah! :-) Okay let's figure this out ... here's an example that I narrowed it down to:

```
function do_stuff(t: table[string] of string): table[string] of string {
        local pending = 0;

        t["foo"] = "bar";

        return when ( pending == 0 )
                {
                return t;
                }
}


event zeek_init() {
        local t: table[string] of string = table();

        when ( local t2 = do_stuff(t) )
                {
                # This won't have the member that do_stuff() added ...
                print(t);
                # ... but this will:
                print(t2);
                }
}
```

This prints:

```
{

}
{
[foo] = bar
}
```

Since the first print doesn't show "foo", I assumed `t` is copied into `do_stuff`?

rsmmr:
Interesting. :-) I believe that's an artifact of different effect: the frame is actually deep-copied multiple times, and the body is getting its own copy, separate from the condition (so do_stuff() operates on a different `t` than `print`). I'm guessing a bit here, but I believe this is the code responsible for that: https://github.com/zeek/zeek/blob/d75663c4b85b88a2c3ab01c7bb0419010192f6ef/src/Trigger.cc#L238

But try this: move the `local t: ...` into `global t: ...`. Now the output is:

```
{
[foo] = bar
}
{
[foo] = bar
}
```

That shows that's it's not `do_stuff` that's copying its parameters: if it did, `t` would no longer link to the global.

rsmmr:
Yes, this all matches what I would expect, and "generally cannot make lasting modifications" seems a fair summary for the doc. :-)

ckreibich:
Okay Robin, take a look. I hope it now explains the high-level takeaway without needing to go into too much detail...

rsmmr (approved):
Looks good!

PR #99, opened 2022-01-03
Zeekygen docs update for cluster controller documentation

ckreibich:
This is the result of a manual `./ci/update-zeekygen-docs.sh` run, to speed things up today.


timwoj (approved):
There's always that one last thing isn't there? 😄

PR #108, opened 2022-03-03
Add note to call out the danger of using "set1 = set1 | set2"

ckreibich:
During some performance measurements I compared this to using "add" and noticed
that this quickly gets very slow as set sizes go up (ballpark: OR-ing a
single-member set onto one with 5000 members is well north of a second).

### DIFF
--- a/script-reference/operators.rst
+++ b/script-reference/operators.rst
@@ -154,6 +154,8 @@ Set operators
   * - Set difference
     - ``s1 - s2``
 
+.. _assignment-operators:
+
 Assignment operators
 --------------------
 
--- a/script-reference/types.rst
+++ b/script-reference/types.rst
@@ -1063,6 +1063,13 @@ Set operations
 You can compute the union, intersection, or difference of two sets
 using the ``|``, ``&``, and ``-`` operators.
 
+.. note::
+
+   Use ``+=`` instead of ``|`` to grow an existing set. That is,
+   say ``s += new_s`` instead of ``s = s | new_s``. The latter requires
+   copying both input sets and thus quickly deteriorates runtime. See
+   :ref:`assignment-operators` for details.
+
 You can compare sets for equality (they have exactly the same elements)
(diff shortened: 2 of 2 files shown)


vpax:
Christian, can you please send me the test script you used for that?  I'd like to explore the issue a bit more, I'm quite surprised it's _that_ bad!

vpax:
Ah - I misunderstood your original comment.  The above makes sense (it's going to have O(N^2) execution) - though is still appalling of course.  It's not so easy to fix in full generality (easy to fix for specific cases), but I'll think about it.

ckreibich:
Cool, yeah. If you compare to using `add` it's pretty striking, so I figured the note has merit.

JustinAzoff:
Are we sure this is a time complexity thing?

```
s = s | set(unique_id(""));
```

has to do effectively

```
news = s | set(unique_id(""));
olds = s;
s = news;
unref(olds);
```

Compared to `add` that can just insert the new element via a mutation.  I'd imagine the copying and unref'ing has the bigger impact.. or at least would assuming the `|` operator is implemented as efficiently as it could be.  What we would really need is an operator like

```
s |= set(unique_id(""));
```
that could be used to do the in place union.

timwoj:
To be fair, `string` has the same sort of issues, but it was never designed to be mutable (although maybe we should revisit that sometime). A `+=` operation for string that didn't require making a completely new string would likely be a pretty big performance gain.

But that's digressing from this PR, sorry 😃

vpax:
Yeah by time complexity I meant that Christian's script is O(N^2).  The operation is O(N) for N = size of the larger set.  So it's the repeated copying that is killing things.

Regarding the need for the operator, it's already there for singletons via `add`.  What I'm currently working on will support:
> s += { any_other_compatible_set };

for efficient additions to set.  We could introduce `|=` for syntactic consistency, but `+=` is already around and already needs to have these semantics, so seems sufficient to me.

rsmmr (approved):

PR #119, opened 2022-06-30
Document the management framework

ckreibich:
I think this is nearing completion — two sections at the end are still missing, as well as architectural diagrams, which I'm hoping to be able to complete fairly quickly tomorrow morning.


ckreibich:
Okay I've marked this as ready. It still needs the diagrams, which I'll tackle in a subsequent PR.

PR #121, opened 2022-07-03
Add CI to zeek-docs and bump pygments

ckreibich:
The merge of the Management framework docs [broke](https://github.com/zeek/zeek/runs/7165231425?check_suite_focus=true) Zeek's `Generate Docs` job but works in RTD, due to a version mismatch in the pygments highlighter. It'd be nice to get a heads-up of such things prior to merging, so this runs the build via `make` in a Github action triggered on PRs.

The setup is the same as in Zeek (treat warnings as errors), but this uses built-in `sphinx-build` features for this instead of buffering the output and grepping it. If folks like this we could adopt this [in the Zeek action](https://github.com/zeek/zeek/blob/master/.github/workflows/generate-docs.yml#L90-L98) too.

To fix the failure itself, I'm bumping pygments so we test with the same version RTD currently uses.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -1,0 +1,17 @@
+name: Generate Documentation
+
+on:
+  pull_request
+
+jobs:
+  generate:
+    if: github.repository == 'zeek/zeek-docs'
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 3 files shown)


0xxon (approved):

timwoj:
I believe the pygments change needs to be brought over into the 5.0 docs branch as well.

PR #126, opened 2022-07-08
Implement merge_domaindata() on Zeek domain to support parallel builds

### LINKED ISSUE #122: Figure out how to make zeek-docs builds run in parallel
It's possible to pass a `-j` argument to `sphinx-build` to make it run tasks in parallel similar to `make`. There's some issues  that crop up with our reader and writer for Zeekygen when we do this though. It'd be good to sort through those issues though since our docs builds take a really long time, and it slows down development quite a bit.

awelzel:
Only the resulting environment.pickle file appears different.

    $ diff -ru ./single-proc-html/ ./build/html/
    Binary files ./single-proc-html/.doctrees/environment.pickle and
    ./build/html/.doctrees/environment.pickle differ
    $

I am not quite sure incremental updates will work right,
but this may be a start.

Changes build time locally from 8m 40s to 2m 50s.

Closes #122, maybe.


awelzel:
@timwoj - this doesn't seem to have much of an effect on the github action timings. Do you see a difference in runtime locally?

timwoj:
On macOS I get the following warning. It's maybe 5 seconds faster on your branch, but that's probably just that it had cached part of it in RAM on the second run.

> For security reason, parallel mode is disabled on macOS and python3.8 and above. For more details, please read https://github.com/sphinx-doc/sphinx/issues/6803

On Linux, it's wildly faster. `master` took 423 seconds and your branch took 122 seconds. That's with 8 CPUs though. If I force it down to 2 CPUs similar to how it is on GitHub, it takes 448 seconds.

awelzel:
The OSX piece is sad. It seems to have been fixed with Sphinx 4.4 though:

https://github.com/sphinx-doc/sphinx/issues/6881#issuecomment-1007952881

> On Linux, it's wildly faster. master took 423 seconds and your branch took 122 seconds. That's with 8 CPUs though. If I force it down to 2 CPUs similar to how it is on GitHub, it takes 448 seconds.

Yep, it doesn't seem `-j` scales quite as nice as maybe expected.

I wonder if caching the build dir could be an alternative, but not sure how safe/reliable that works.

timwoj:
I installed a newer version of sphinx into a virtualenv on macOS here and can confirm that it seems they fixed parallel builds on here now. This is what I have in requirements.txt now:

```
#jinja2<3.1
#pygments>=2.12.0
#docutils==0.16
sphinx_rtd_theme==1.0.0
Sphinx==5.0.2
GitPython==3.1.27
```

`docutils` didn't get installed at all. jinja2 3.1.2 and pygments 2.12.0 got automatically installed. It seems that the newer version of Sphinx also fixed whatever conflict they were having with the newer version of jinja2. I was also able to bump the sphinx_rtd_theme package up to the latest version.

awelzel:
After writing some documentation and building it quite regularly and being annoyed about the lengthy process, I included the version updates and cleaned up the `-j` passing a bit in the Makefile. Otherwise no changes. I skimmed the produced docs and they seemed okay.

> docutils didn't get installed at all

It installed for me as `0.17.1`, so I kept that in the requirements.txt.

Should we merge this and see if something breaks? :man_shrugging:

timwoj:
Sounds good to me.

Makefile
```diff
@@ -11,7 +13,7 @@ clean:
 	rm -rf build/html
 
 html: builddir
-	sphinx-build -b html $(SPHINXOPTS) . ./build/html
+	sphinx-build -j $(NUMJOBS) -b html $(SPHINXOPTS) . ./build/html
```
timwoj, line 16: It looks like as of sphinx 1.7 you can just pass `-j auto` and it'll figure it out for you: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-j

awelzel: Ah, thanks. I did use `NUMJOBS=4 make` or so a few times locally to leave some cycles for the rest of the system, but can achieve that in other ways, too. Let me squelch it with a separate commit and if we want to have it back can revert it.

timwoj: In that case, maybe do `NUMJOBS ?= "auto"`. Then it could use the variable if you've set it.

timwoj (approved):
We can look at the RTD thing separately/

PR #137, opened 2022-08-22
statements: Leave a note for index, value iteration of vectors

awelzel:
I'm not 100% certain if the "core exposes value variables for free" also
applies to the vectors or is more a table/HashKey thing. Happy to
re-structure if the current version isn't quite right.


ckreibich:
I think it holds and is fine to keep: for tables this likely refers to the fact that for a lookup the work is all in locating the key's corresponding `DictEntry`. Once located, the latter points directly at the value, which can then be passed on to the script layer. For vectors this is also true — the stored element gets converted from a `ZVal`, but that is either very cheap or just a pointer assignment.

ckreibich (approved):

Issue #91, opened 2021-08-25
Expand File Analysis Framework docs


ckreibich:
Aside from covering the protocols the framework supports (#62) we should also expand coverage of the framework itself. For example:
- cover unmentioned events (`file_sniff` isn't mentioned)
- explain disable_analyzer() vs add_analyzer()/remove_analyzer() (in particular, for the latter the ability to add/remove repeatedly and independently depending on the provided events)
- explain the relationship of files to connections
- explain chunks vs streams

PR #139, opened 2022-08-25
List protocols supporting file analysis in the FAF docs

### LINKED ISSUE #62: Explain protocol support in File Analysis Framework
The [FAF docs](https://docs.zeek.org/en/master/frameworks/file-analysis.html) currently focus on development/application of the framework. We should say more about its use, including which protocols actually support FAF at the moment. Could be in the FAF section, or via a new section about file extraction.

ckreibich:
Closes #62.


frameworks/file-analysis.rst
```diff
@@ -23,6 +23,19 @@ File Analysis
     provide analysis specifically for files that is analogous to the
     analysis Zeek provides for network connections.
 
+Supported Protocols
```
timwoj, line 26: Should this mention that these are the protocols supported by default but not necessarily all of them (i.e. packages)?

timwoj (changes requested):

frameworks/file-analysis.rst
ckreibich: Yeah, agree. Will do. I'm on #91 right now where I've state the same thing for the file analyzes themselves (like entropy, PE, etc).

timwoj (approved):

PR #140, opened 2022-08-26
Edit pass over installation page

### LINKED ISSUE #134: advertise docker as installation method
Hey guys,
* At first, I was surprised to not find docker as an advertised method for installation in https://docs.zeek.org/en/master/install.html ...
* Then I found several unofficial zeek images in dockerhub ... 
* Then I went in your CI logs at the main repo and saw that you do have jobs to build and publish official images which finally led me to what I believe is the official repo (is it??): https://hub.docker.com/r/zeekurity/zeek

To me it seems a no brainer to advertise the dockerhub link in the install page, perhaps with one or two examples of how to run it... You do all the hard work of building the container images and then it's not in the docs where one would expect it to find it! 😅  even just a one liner with a link would be good enough!

ckreibich:
This adds coverage of our Docker images and restructures the section to cover dependencies only in the building-from-sources section. It now goes from easiest to most involved: Docker -> binary packages -> build from source. I've also updated some technical details that looked out of date.

This doesn't yet include AWS ECR instructions. We can add these when our images are up and running there.

RTD preview: https://docs.zeek.org/en/topic-christian-gh-134-installation/install.html

Closes #134.

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -18,49 +18,143 @@
 Installing Zeek
 ===============
 
-Prerequisites
+To run Zeek, grab our official Docker images, download our Linux binary
+packages, install via Homebrew on your Mac, use the ports collection on FreeBSD,
+or build Zeek yourself.
+
+Docker Images
 =============
 
-Before installing Zeek, you'll need to ensure that some dependencies
-are in place.
+We provide official Docker images on Docker Hub at https://hub.docker.com/u/zeekurity.
+
+    * For the latest feature release: ``docker pull zeekurity/zeek:latest``
+    * For the latest LTS release: ``docker pull zeekurity/zeek:lts``
+    * For a specific release: ``docker pull zeekurity/zeek:5.0.0-rc1``
+    * For the nightly build: ``docker pull zeekurity/zeek-dev:latest``
+
+The images run Debian and feature a complete Zeek installation with ``zeek``,
+``zkg``, and the Spicy toolchain, but are otherwise minimal to avoid bloat in
+derived images. For example, if you'd like to install Zeek plugins in those
+images, you'll need to install their needed toolchain, typically at least
+``g++`` for compilation, ``cmake`` and ``make`` as build tools, and
+``libpcap-dev`` to build against Zeek headers. Similarly, you'll need ``g++``
+for Spicy's JIT compilation, as well as ``cmake`` and ``make`` to build Spicy
+analyzer packages.
+
+  .. code-block:: console
+
+    apt-get update
+    apt-get install -y --no-install-recommends g++ cmake make libpcap-dev
+
+The Dockerfile lives `here <https://github.com/zeek/zeek/blob/master/docker/Dockerfile>`_.
+
+Binary Packages
+===============
+
+Linux
+-----
+
+We provide `binary packages <https://build.opensuse.org/project/show/security:zeek>`_
+for a wide range of Linux distributions via the `openSUSE Build Service
+<https://build.opensuse.org/>`_. To install, first add the relevant OBS
+package repository to your system, then use your system's package manager
+as usual.
+
+We provide packages for:
+
+    * the `latest Zeek feature release <https://software.opensuse.org/download.html?project=security%3Azeek&package=zeek>`_ (`sources <https://build.opensuse.org/package/show/security:zeek/zeek>`__)
+    * the `latest Zeek LTS release <https://software.opensuse.org/download.html?project=security%3Azeek&package=zeek-lts>`_ (`sources <https://build.opensuse.org/package/show/security:zeek/zeek-lts>`__)
+    * `nightly builds <https://software.opensuse.org/download.html?project=security%3Azeek&package=zeek-nightly>`_ (`sources <https://build.opensuse.org/package/show/security:zeek/zeek-nightly>`__)
+    * `release candidates <https://software.opensuse.org/download.html?project=security%3Azeek&package=zeek-rc>`_ (`sources <https://build.opensuse.org/package/show/security:zeek/zeek-rc>`__)
+
+For example, for the latest feature release on CentOS Stream 8 you'd run the following as root:
+
+  .. code-block:: console
+
+     cd /etc/yum.repos.d/
+     wget https://download.opensuse.org/repositories/security:/zeek/CentOS_8_Stream/security:zeek.repo
+     dnf install zeek
+
+For the LTS release on Ubuntu 22.04 the steps look as follows:
+
+  .. code-block:: console
+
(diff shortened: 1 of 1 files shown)


install.rst
```diff
@@ -225,13 +252,16 @@ Git repositories hosted at https://github.com/zeek:
...
 
 .. note:: If you choose to clone the ``zeek`` repository
    non-recursively for a "minimal Zeek experience", be aware that
    compiling it depends on several of the other submodules as well, so
    you'll likely have to build/install those independently first.
 
+Configuring and Building
+------------------------
+
 The typical way to build and install from source is (for more options,
```
timwoj, line 265: You can remove the `for more options` sidebar here. It's mentioned later already.

timwoj (changes requested):
One minor nit, looks fine otherwise. 

I had some issues with one of the command lines for installing dependencies recently but I can't remember what system it was on nor what the problem was. If it comes back up, I'll make a note of it.

install.rst
```diff
@@ -18,49 +18,136 @@
...
+To run Zeek, grab our official Docker images, download our Linux binary
+packages, install via Homebrew on your Mac, use the ports collection on FreeBSD,
+or build Zeek yourself.
+
+Docker Images
 =============
 
-Before installing Zeek, you'll need to ensure that some dependencies
-are in place.
+We provide `official Docker images <https://hub.docker.com/u/zeekurity>`_ on the
```
awelzel, line 28:
```suggestion
We provide `official Docker images <https://hub.docker.com/u/zeekurity>`_ on
```

Also wonder if `zeekurity` should shine through here already than being hidden behind "official Docker images" link. While it may be obvious to the initiated, "zeekurity" is only later mentioned in the pull commands, but otherwise doesn't seem to be referenced in the in the zeek-docs tree. For someone new, the username/repo name may be surprising (much lesser if it was just `zeek`, but that seems taken).

Maybe something like, "we provide official docker images in the zeekurity/zeek repo on Docker Hub..." ?

install.rst
```diff
@@ -18,49 +18,136 @@
...
+Docker Hub. These images are based on Debian 11 and feature a complete Zeek
```
awelzel, line 29: That `11` will need updating often, would just mentioning Debian be enough?

install.rst
```diff
@@ -18,49 +18,136 @@
...
+installation with ``zeek``, ``zkg``, and the Spicy toolchain, but are otherwise
+minimal to avoid bloat in derived images. For example, if you'd like to install
+Zeek plugins in those images, you'll need to install the needed compiler
+toolchain.
```
awelzel, line 33: The Spicy toolchain has limited functionality without a compiler, too, right? Is it worth mentioning it prominently if the interesting functionality doesn't quite work out of the box?

install.rst
```diff
@@ -18,49 +18,136 @@
...
-    * CMake 3.15 or greater              (https://www.cmake.org)
-    * Make
+Linux
+-----
+
+We provide `binary packages <https://build.opensuse.org/project/show/security:zeek>`_
+for a wide range of Linux distributions via the `openSUSE Build Service
+<https://build.opensuse.org/>`_. To install, first add the relevant OBS
+package repository to your system, then use your system's package manager
+as usual. Our packages resolve third-party dependencies for you.
```
awelzel, line 54:
```suggestion
as usual.
```
Might be opinionated, but seems the contrary would be worth mentioning rather than the expected behavior.

awelzel (approved):
Nit bits.

install.rst
ckreibich: @bbannier, how do you think we should word this?

install.rst
ckreibich: Done

install.rst
ckreibich: Done

install.rst
ckreibich: Done

install.rst
ckreibich: Done

install.rst
bbannier: > @bbannier, how do you think we should word this?

We briefly had talked about this previously. Most of the time one just needs to install a handful of packets, but for wildly different reasons:
- `g++` is needed for Spicy JIT
- `cmake`, `make`: typical requirement of Spicy analyzer packages, e.g., the one generated with the default template
- `libpcap-dev` (provided by `libpcap0.8-dev`): dev dependency required to build with certain Zeek headers, e.g., `packet_analysis/Analyzer.h`.

When we last discussed this there was some preference for adding a script to the image setting up this toolchain, but looking back this feels too intrasparent to me given that it might not include the actual requirements of a given use case, so maybe it would be better to give a short explicit blurb in the docs, e.g.,

> We provide official Docker images on Docker Hub at https://hub.docker.com/u/zeekurity.
> The images run Debian and feature a complete Zeek installation with ``zeek``,
> ``zkg``, and the Spicy toolchain, but are otherwise minimal to avoid bloat in
> derived images. For example, if you'd like to install Zeek plugins in those
> images, you'll need to install the needed toolchain, typically `g++` for
> compilation, `cmake` and `make` as a build tool, and `libpcap-dev` to build
> against Zeek headers:
>
>     apt-get update
>     apt-get install --no-recommends g++ cmake make libpcap-dev

install.rst
```diff
@@ -18,49 +18,135 @@
...
+To run Zeek, grab our official Docker images, download our Linux binary
+packages, install via Homebrew on your Mac, use the ports collection on FreeBSD,
+or build Zeek yourself.
+
+Docker Images
 =============
 
-Before installing Zeek, you'll need to ensure that some dependencies
-are in place.
+We provide official Docker images on Docker Hub at https://hub.docker.com/u/zeekurity.
```
bbannier, line 28: Just wondering, I thought there was some work on the way to publish images to the ECR as well (https://gallery.ecr.aws/?searchTerm=zeek), but I cannot find the ticket.

install.rst
ckreibich: Great, thanks. I recall the script idea and still like it, fwiw. I'll word in as you suggested.

install.rst
ckreibich: Yes that's stalled on me, I've been meaning to explore CLI access to the AWS ECR to sync over the existing images. I don't think there's currently a ticket.

timwoj (approved):

install.rst
bbannier: Not sure about CLI, but publishing from CI is pretty simple with the actions we already use (just requires an additional step to log in at ECR, and another remote to push to); the Spicy setup which already does it is [here](https://github.com/zeek/spicy/blob/main/.github/workflows/docker.yml).

PR #141, opened 2022-08-29
cluster-setup: Revamp Netmap section

awelzel:
I'm not sure this is too verbose, but given that generally seemed one
needs to find small pieces of info in various places, I'm not feeling too
bad putting a few details in.

This is mostly from a "running without zeekctl" perspective.

### DIFF
--- a/cluster-setup.rst
+++ b/cluster-setup.rst
@@ -142,9 +142,102 @@ PF_RING, see the documentation on :ref:`how to configure Zeek with PF_RING
 Netmap
 ^^^^^^
 
-FreeBSD has an in-progress project named Netmap which will enable flow-based
-load balancing as well.  When it becomes viable for real world use, this
-document will be updated.
+`Netmap <https://github.com/luigirizzo/netmap>`_ is a framework for fast
+packet I/O that is natively supported on FreeBSD since version 10.
+On Linux it can be installed as an out-of-tree kernel module.
+
+FreeBSD
+"""""""
+FreeBSD's libpcap library supports netmap natively. This allows to prefix
+interface names with ``netmap:`` to instruct libpcap to open the interface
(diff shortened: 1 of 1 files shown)


ckreibich:
Thanks Arne, this is great. How about two subsections fro this, one for FreeBSD, one for Linux? The way the section starts out, it sounds like it's completely about FreeBSD.

> This is mostly from a "running without zeekctl" perspective.

Yes, great — we should no longer add material that specifically targets the `zeekctl` scenario.

cluster-setup.rst
```diff
@@ -142,9 +142,88 @@ PF_RING, see the documentation on :ref:`how to configure Zeek with PF_RING
...
+The plugin can also be used on FreeBSD in case libpcap's native support is found
+to be insufficient. The plugin further provides basic zeekctl support.
+
+.. warning::
+
+  If you're using the zeek-netmap plugin on FreeBSD, the interface specification
+  given to Zeek needs to change from `netmap:zeek}0/x` to `netmap::zeek}0/x` - a single colon more.
+  In the first case, Zeek instantiates the libpcap packet source and passes `netmap:zeek}0`
+  as parameter. In the second case, `netmap::` is interpreted by Zeek, it
+  instantiates teh netmap packet source `zeek}0/x` is treated as interfac
```
ckreibich, line 224: Typos. :-)

ckreibich (changes requested):
Looks great, just requesting changes for the typos and separating FreeBSD and Linux, if you're up for that.

awelzel:
> Looks great, just requesting changes for the typos and separating FreeBSD and Linux, if you're up for that.

Done. It's still fairly FreeBSD heavy, but think this mostly comes from being much more accessible there as compared to Linux. Also not sure we should recommend netmap on Linux..

ckreibich (approved):

PR #146, opened 2022-09-20
cluster-setup: Stab at AF_PACKET support documentation

### LINKED ISSUE #87: Add modern af_packet cluster setup example
The current install docs describe how to use pf_ring, which was the only reliable way to run a cluster many years ago.

pf_ring still works, but unless the extra features are required, af_packet should be easier to get started with and a good choice for most people.

awelzel:
...related to zeek/zeek#2401.

Closes #87

### DIFF
--- a/cluster-setup.rst
+++ b/cluster-setup.rst
@@ -144,8 +144,9 @@ AF_PACKET
 ^^^^^^^^^
 
 On Linux, Zeek supports `AF_PACKET sockets <https://docs.kernel.org/networking/packet_mmap.html>`_ natively.
-Currently, this is provided by including the `external zeek-af_packet-plugin <https://github.com/J-Gras/zeek-af_packet-plugin>`_
-in default builds of Zeek for Linux.
+Currently, this is provided by including the `external Zeek::AF_Packet plugin <https://github.com/J-Gras/zeek-af_packet-plugin>`_
+in default builds of Zeek for Linux. Additional information can be found in
+the project's README file.
 
 To check the availability of the ``af_packet`` packet source, print its information using ``zeek -N``::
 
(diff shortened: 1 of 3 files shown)


ckreibich:
Thank you Arne! @J-Gras would like to look this over?

J-Gras (commented):
This starts with a nice introduction to AF_Packet. Beginning at the paragraph about the fanout id, the documentation gets quite complex, dealing with some special use case. From a user's perspective I would be missing an example of a more common setup using zeekctl. Furthermore, notes on privileges, promiscuous mode and offloading might be very valuable as well. I would suggest to focus on a simple, common use case (maybe just copy from the plugin readme) and the most common pitfalls. Special cases and tuning options could be discussed subsequently or moved to the plugin readme.

J-Gras:
@initconf might be able to provide feedback as well.

cluster-setup.rst
```diff
@@ -139,6 +139,77 @@ threaded and can't natively utilize all of the cores.  If you want to use
...
+
+
+For the most basic usage, prefix the interface with ``af_packet::`` when invoking Zeek::
+
+    zeek -i af_packet::eth0
+
+The more interesting use-case is to use AF_PACKET to run multiple Zeek workers
+and have their packet sockets join what is called a fanout group.
+In such a setup, network packets are load-balanced across Zeek workers.
+By default load balancing is based on flow-hashes.
```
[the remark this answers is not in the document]
awelzel: Added.

cluster-setup.rst
```diff
@@ -139,6 +139,100 @@ threaded and can't natively utilize all of the cores.  If you want to use
...
+and have their packet sockets join what is called a fanout group.
+In such a setup, the network traffic is load-balanced across Zeek workers.
+By default load balancing is based on symmetric flow hashes [#]_.
+
+For example, running two Zeek workers listening on the same network interface,
+each worker analyzing approximately half of the network traffic, can be done
+as follows::
+
+    zeek -p eth0-worker-01 -i af_packet::eth0 &
+    zeek -p eth0-worker-02 -i af_packet::eth0 &
```
rsmmr, line 208: What's the prefixes for here? Are they needed?

awelzel: Not needed - they are leftovers from a previous version. Will remove.

ckreibich (approved):

ckreibich:
I merged this to accompany zeek/zeek#2401. I think we'll have to see about zeekctl-specific docs after all ... there was sufficient interest in the AF_PACKET inclusion on Slack that I'm tempted to say people will want to use this right away, and so the docs should still cover it. But for now this is a great step forward and we can revisit as 5.2 approaches.

PR #150, opened 2022-10-19
Add note on &is_used for events.

J-Gras:
Just checked that this works via https://try.zeek.org/#/tryzeek/saved/612619

### DIFF
--- a/script-reference/types.rst
+++ b/script-reference/types.rst
@@ -1649,6 +1649,10 @@ from the canonical prototype.
 To use :zeek:attr:`&default` on event arguments, it must appear on the
 first, canonical prototype.
 
+Employing its static analysis capabilities, Zeek will warn if it cannot
+determine that an event will ever be triggered. In case the warning is not
+appropriate (e.g., the event might be triggered remotely via broker),
+:zeek:attr:`&is_used` can be applied to suppress the warning.
 
 .. zeek:native-type:: hook
 


awelzel:
Thanks! Seems the documentation around `&is_used` is a bit stale. I think it applies/works for non-exported functions, too.

Maybe extending the dedicated `&is_used` section in `scripting-reference/attributes.rst` would be an alternative place?

J-Gras:
> Thanks! Seems the documentation around `&is_used` is a bit stale. I think it applies/works for non-exported functions, too.

Out of curiosity: Is there a way to call a function that isn't caught by Zeek?
 
> Maybe extending the dedicated `&is_used` section in `scripting-reference/attributes.rst` would be an alternative place?

Having it in both places makes sense.

awelzel:
> Out of curiosity: Is there a way to call a function that isn't caught by Zeek?

Suspect generally if you do `@if` things in cluster scenarios some functions may not be used on certain types of nodes (and putting the right `@if` around the helpers may be too burdensome).

```
module MyModule;


function helper() {
  print "helper";
}


@if ( Cluster::local_node_type() == Cluster::MANAGER )
event zeek_init() {
        helper();
}
@endif
```

ckreibich:
> Having it in both places makes sense.

There's also https://docs.zeek.org/en/master/scripting/usage.html, so three total. It'd be great if you could make them all link to https://docs.zeek.org/en/master/script-reference/attributes.html#is-used and cover it in most detail there, while you're at it. (`&is_assigned` is similar, if you're up for it.) Thank you!

vpax:
> Suspect generally if you do `@if` things in cluster scenarios

Yep, that's a common one ...

ckreibich:
@J-Gras are you up for the above change? I can also take it from here if you're busy.

J-Gras:
> @J-Gras are you up for the above change? I can also take it from here if you're busy.

I put it on my list :)

PR #152, opened 2022-10-25
Add &group attribute reference

awelzel:
This is rather sparse currently, we need to have the zeek:attr &group in place for the build to work.


awelzel:
I suspect the build will fail due to the BIFs not yet available. Hoping for @rsmmr to straighten things out.

PR #153, opened 2022-10-27
Clarify the relationship of pattern type syntax to that of flex

### LINKED ISSUE #149: Pattern not support `{-}` and `{+}` operators
According to the [docs](https://docs.zeek.org/en/master/script-reference/types.html#type-pattern), Zeek accepts “patterns supported by the [flex lexical analyzer](http://westes.github.io/flex/manual/Patterns.html).”  However, character class unions/differences are not working. 

Running the following test-script with Zeek 5.0: 
```
const mypattern: pattern = /([[:print:]]{-}[[:alnum:]]foo)/;
```
Get the following error:
```
error in ././try.zeek, line 1: error compiling pattern /^?(.|\n)*(([[:print:]]{-}[[:alnum:]]foo))/
```

Reference to report on slack : https://zeekorg.slack.com/archives/CSZBXF6TH/p1662645132381609

ckreibich:
Let me know if this suits.

Resolves #149.

### DIFF
--- a/script-reference/types.rst
+++ b/script-reference/types.rst
@@ -520,15 +520,168 @@ Type Conversions
 pattern
 -------
 
-A type representing regular-expression patterns that can be used
-for fast text-searching operations.  Pattern constants are created
-by enclosing text within forward slashes (``/``) and use the same syntax
-as the patterns supported by the `flex lexical analyzer
-<http://westes.github.io/flex/manual/Patterns.html>`_.  The speed of
-regular expression matching does not depend on the complexity or
-size of the patterns.  Patterns support two types of matching, exact
-and embedded.
+A type representing regular-expression patterns that can be used for fast
+text-searching operations.  Pattern constants are created by enclosing text
+within forward slashes (``/``) and support a large subset of the `flex lexical
+analyzer <http://westes.github.io/flex/manual/Patterns.html>`_ syntax.  As in
+other implementations, patterns consist of ordinary and special characters.
+Patterns such as ``/a/`` or ``/A0123/`` match that specific character byte or
+character sequence, case-sensitively. Special characters and modifiers customize
+matching, as follows:
 
+.. list-table::
+  :header-rows: 1
+
+  * - Syntax
+    - Meaning
+
+  * - ``^``
+    - Matches the beginning of the input.
+
+  * - ``$``
+    - Matches the end of the input.
+
+  * - ``.``
+    - Matches any character except newline (but see the ``(?s:`` and ``/s`` modifiers).
+
+  * - ``<expr>*``
+    - Matches zero or more instances of ``expr``.
+
+  * - ``<expr>+``
+    - Matches one or more instances of ``expr``.
+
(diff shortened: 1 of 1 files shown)


rsmmr (approved):
This probably suggests a follow up item of documenting our syntax, given that we can't just refer to flex anymore.

ckreibich:
Yeah good point Robin. I went ahead and did that in this PR still, take a look. Hope I covered it all — I intentionally skipped the `{NAME}` feature since I don't think it's actually usable in practice and it looks like we added it in the grammar in case we use it in the future.

script-reference/types.rst
```diff
@@ -520,15 +520,109 @@ Type Conversions
...
+    - Matches ``expr`` ``n`` or more times, where ``n`` is a non-negative integer.
+
+  * - ``<expr>{n,m}``
+    - Matches ``expr`` between ``n`` and ``m`` times, inclusively, where ``n``
+      and ``m`` are non-negative integers and ``m >= n``.
+
+  * - ``<expr1>|<expr2>``
+    - Matches either ``expr1`` or ``expr2``. In contrast to the above repetition
+      operators, expressions here span multiple characters: in ``ab+`` the ``+``
+      applies to the ``b``, while ``ab|cd`` matches ``"ab"`` or ``"cd"``.
```
rsmmr, line 569: I'd actually skip the 2nd sentence, find that more confusing than helpful.

```suggestion
    - Matches either ``expr1`` or ``expr2``. 
```

script-reference/types.rst
```diff
@@ -520,15 +520,109 @@ Type Conversions
...
+  * - ``(<expr>)``
+    - Groups the contained expression to form a building-block of a more complex
+      one.
+
+  * - ``"<chars>"``
+    - Matches literal strings, without the quotation marks. These always match
+      as given, even if case-insensitivity is active.
+
+  * - ``[<chars>]``
+    - Defines a character class, matching any of the contained characters.
```
rsmmr, line 580: Let's add the character classes as well. Suggestion:

In addition to literal characters, the following character classes as supported as well: `[:alnum:]`,`[:alpha:]`,`[:blank:]`,`[:cntrl:]`,`[:digit:]`,`[:digit:]`,`[:graph:]`,`[:lower:]`,`[:print:]`,`[:punct:]`,`[:space:]`,`[:upper:]`,`[:xdigit:]`


They come with their standard POSIX meaning, see for example [this reference](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX_Basic_Regular_Expressions#Character_classes>).

rsmmr (approved):

script-reference/types.rst
ckreibich: I went back and forth over that one a bunch. :-) Happy to remove it.

script-reference/types.rst
ckreibich: Oh right, oversight on my part! I'll work it in.

PR #157, opened 2022-11-11
Add documentation for building Zeek on Windows

timwoj:


install.rst
```diff
@@ -201,6 +202,51 @@ To install these, you can use:
...
+* Windows
+
+  Compiling on Windows requires the installation of a development environment.
+  Zeek currently builds on Visual Studio 2019, and you can either install the
+  full version including the UI tools or you can install the command-line tools
+  and build from a shell. The instructions below describe how to install the
+  command-line tools, but are not necessary if you install the full VS2019
+  package. You will need to install Chocolatey_ in order to install the
+  dependencies as instructed below. It's possible to install them from other
+  sources, but that is left to the reader to determine how.
```
ckreibich, line 214:
```suggestion
  sources, which we leave to the reader.
```
Could perhaps also add that feedback on such alternatives is welcome?

install.rst
```diff
@@ -201,6 +202,51 @@ To install these, you can use:
   ``swig``, ``swig-python``, ``openssl``, ``bison``, and ``flex`` packages
   provide the required dependencies.
 
+* Windows
+
+  Compiling on Windows requires the installation of a development environment.
+  Zeek currently builds on Visual Studio 2019, and you can either install the
```
ckreibich, line 208: I'm a n00b here — perhaps a link to an installation entry point would help? Feel free to ignore if that's too obvious for Windows folk.

install.rst
```diff
@@ -201,6 +202,51 @@ To install these, you can use:
...
+
+     c:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat x86_amd64
+
+  Now you can build via cmake:
+
+  .. code-block:: console
+
+     mkdir build
+     cd build
+     cmake.exe .. -DCMAKE_BUILD_TYPE=release -DENABLE_ZEEK_UNIT_TESTS=yes -G Ninja
```
ckreibich, line 243: Where do cmake and Ninja come from, btw? Visual Studio?

install.rst
```diff
@@ -201,6 +202,51 @@ To install these, you can use:
   ``swig``, ``swig-python``, ``openssl``, ``bison``, and ``flex`` packages
   provide the required dependencies.
 
+* Windows
+
```
ckreibich, line 206: Do you think we should add a note here to indicate that all of this is experimental? I know you put it in `NEWS` ... I'm just thinking that new folks might bump into this section more readily than the NEWS file.

ckreibich (approved):

install.rst
timwoj: Added mention of msys2 and cygwin.

install.rst
timwoj: Yep, they're part of that installation. Calling that `vcvarsall` script adds them your path.

install.rst
timwoj: Added a section at the top about it.

PR #163, opened 2022-12-07
Suggest installing recommended packages for container image.

bbannier:
When documenting additional packages needed to get Spicy running in our container image, we recommended a command line which used `--no-install-recommends`. This command line doesn't seem to work as some recommended depedency does not get installed.

With this patch we now do not document skipping recommended packages anymore.

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -52,7 +52,7 @@ analyzer packages.
   .. code-block:: console
 
     apt-get update
-    apt-get install -y --no-install-recommends g++ cmake make libpcap-dev
+    apt-get install -y g++ cmake make libpcap-dev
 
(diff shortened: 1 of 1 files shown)


awelzel:
As mentioned in the [zeek.org Slack](https://zeekorg.slack.com/archives/C012CQV1K29/p1670426402022799?thread_ts=1670414320.379089&cid=C012CQV1K29) adding the actual requirement should be preferred.


> This command line doesn't seem to work as some recommended depedency does not get installed.

`--no-install-recommends` stop installing recommended packages. If there's a package that is actually required, IMO that's actually a required dependency, not something recommended.

The `Recommends` fields of Debian/Ubuntu packages may also add/remove things we don't/do need outside of our control.

PR #167, opened 2023-01-06
Expand file analysis framework docs

ckreibich:
This separates general file analyzer attachment/removal concepts from subsequent examples, and expands the set of examples to include custom content analysis via the data event file analyzer. Also includes various small tweaks for better flow.

### DIFF
--- a/frameworks/file-analysis.rst
+++ b/frameworks/file-analysis.rst
@@ -90,12 +90,15 @@ Adding Analysis
 ===============
 
 Zeek supports customized file analysis via `file analyzers` that users can
-attach to observed files.  Once attached, file analyzers start receiving the
-contents of files as Zeek extracts them from ongoing network connections.
+attach to observed files. You can attach analyzers selectively to individual
+files, or register them for auto-attachment under certain conditions. Once
+attached, file analyzers start receiving the contents of files as Zeek parses
+them from ongoing network connections.
+
 Zeek comes with the following built-in analyzers:
 
     * :ref:`plugin-zeek-filedataevent` to access file content via
-      events (as sequential data streams or non-sequential content chunks),
+      events (as data streams or content chunks),
     * :ref:`plugin-zeek-fileentropy` to compute various entropy for a file,
     * :ref:`plugin-zeek-fileextract` to extract files to disk,
     * :ref:`plugin-zeek-filehash` to produce common hash values for files,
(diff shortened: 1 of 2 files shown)


frameworks/file-analysis.rst
```diff
@@ -90,34 +90,69 @@ Adding Analysis
 ===============
 
 Zeek supports customized file analysis via `file analyzers` that users can
-attach to observed files.  Once attached, file analyzers start receiving the
-contents of files as Zeek extracts them from ongoing network connections.
+attach to observed files. You can attach analyzers selectively to individual
+files, or register them for auto-attachment under certain conditions. Once
+attached, file analyzers start receiving the contents of files as Zeek extracts
```
awelzel, line 95: Could we avoid "extract" here steer clear of the confusion with the file-extraction part implying persistent storage? Maybe "observe" as you used in the file extraction example further down?

```suggestion
attached, file analyzers start receiving the contents of files as Zeek observes them on the network.
```

frameworks/file-analysis.rst
```diff
@@ -126,38 +161,77 @@ calculate the MD5 of plain text files:
    new file, FakNcS1Jfe01uljb3
    file_hash, FakNcS1Jfe01uljb3, md5, 397168fd09991a0e712254df7bc639ac
 
-Some file analyzers have tunable parameters that need to be
-specified in the call to :zeek:see:`Files::add_analyzer`:
+File extraction
+^^^^^^^^^^^^^^^
+
+The following example sets up simplistic extraction of observed files to
```
awelzel, line 167:
```suggestion
The following example sets up extraction of observed files to
```

frameworks/file-analysis.rst
```diff
@@ -126,38 +161,77 @@ calculate the MD5 of plain text files:
...
+enumerated ``file-NNNN`` strings.
+
+In a production setting you'll likely want to include additional information in
+the output, for example from state attached to the provided file record. The
+Zeek distribution ships with a starting point for such approaches: the
+``policy/frameworks/files/extract-all-files.zeek`` script. For additional
+configurability, take a look at the `file-extraction
+<https://github.com/hosom/file-extraction>`_ Zeek package.
+
+Content analysis
```
awelzel, line 192: Seems Script-Level would be good in the heading? Maybe "Script-Layer File Data Access" as the first sentence indicates?

frameworks/file-analysis.rst
```diff
@@ -126,38 +161,77 @@ calculate the MD5 of plain text files:
...
+        }
+
+    event file_state_remove(f: fa_file)
+        {
+        if ( f$id in hashstate )
+            {
+            print(sha256_hash_finish(hashstate[f$id]));
+            delete hashstate[f$id];
+            }
+        }
```
awelzel, line 234: There have been some perf issues with ANALZYER_DATA_EVENT in production systems where the analyzer wasn't registered. Think multiple GB files triggering lots of events (and maybe only the first few kb are actually of interest).

It is also fairly dangerous to start buffering the data in script-land just because one can.

Not sure these need to be pointed out, but leaving the comment anyway.

awelzel (approved):
Just nits.

frameworks/file-analysis.rst
ckreibich: I wondered about that too. :-) I initially used "parsed" and then reverted it, trying to minimize changes. I'll tweak it.

frameworks/file-analysis.rst
ckreibich: Good one, yeah.

frameworks/file-analysis.rst
ckreibich: Ah, right — thank you for flagging that. I'll add that.

I'll also add a bit about the fact that protocol analyzers can redef additional state onto the file record, like in `base/protocols/http/entities.zeek`.

PR #168, opened 2023-01-09
updated hyperlink to external SMB resource

anthonykasza:
401trg has moved the external resource to GitHub. This change updates the smb.rst file to include the new hyperlink.

### DIFF
--- a/logs/smb.rst
+++ b/logs/smb.rst
@@ -29,7 +29,7 @@ compromised Windows environment.
 
 The paper is available here:
 
-https://401trg.com/an-introduction-to-smb-for-network-security-analysts/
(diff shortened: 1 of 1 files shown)


awelzel:
Thanks. @ckreibich / @timwoj  - should this get backported to `release/5.{0,1}` ?

ckreibich:
Yeah, that'd be great. You can either cherry-pick the commit into the release branches, or (likely better) simply add it to the collection in zeek/zeek#724 because it makes it easier to track what's getting backported. Also, no harm in just adding to the latter if you feel something should be packported — we can always reconsider as we work through the list. Thanks!

PR #175, opened 2023-02-22
install: Point to ./docker directory instead of Dockerfile

awelzel:
Closes zeek/zeek#2805.

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -54,7 +54,8 @@ analyzer packages.
     apt-get update
     apt-get install -y --no-install-recommends g++ cmake make libpcap-dev
 
-The Dockerfile lives `here <https://github.com/zeek/zeek/blob/master/docker/Dockerfile>`_.
+The source files used to create the container images are on
+`GitHub <https://github.com/zeek/zeek/blob/master/docker>`_.
 
 Binary Packages
 ===============


awelzel:
@bbannier also brought up the idea of putting back the old Dockerfile, though that would mostly be an example rather than what's actually being used to create the final container images.

timwoj:
> @bbannier also brought up the idea of putting back the old Dockerfile, though that would mostly be an example rather than what's actually being used to create the final container images.

I think putting the Dockerfile in there would cause confusion about what it's used for.

timwoj (approved):

awelzel:
I've made it a bit more verbose, but essentially the same.

bbannier:
> @bbannier also brought up the idea of putting back the old Dockerfile, though that would mostly be an example rather than what's actually being used to create the final container images.

To be fair, I never suggested to put back dead unused code, but instead was wondering whether this would be another good reason to bring back a more simple `Dockerfile` (the other being that this would simplify the setup a lot where it currently requires careful interaction between Docker setups in multiple files, and a `Makefile`). AFAICT one reason for the current setup was the desire to mount in a local cache during `docker run` executing the Zeek build, but this should also be possible [directly in the `Dockerfile`](https://docs.docker.com/engine/reference/builder/#run---mount).

awelzel:
Thanks @bbannier - I've opened https://github.com/zeek/zeek/issues/2810, but for now I'll move forward fixing the link.

awelzel:
I've also pushed this into release/5.2.

PR #177, opened 2023-03-02
broker: Proxies aren't actually optional

awelzel:
Skimming the base scripts for proxy_topic or proxy_pool, it seems specifically base/frameworks/software uses proxy_pool unconditionally without a fallback.

I introduced that sentence recently, I'm not sure if I had picked that up elsewhere, or just assumed because mostly things work without proxies and there's no warning or so (which maybe we should add).

### DIFF
--- a/frameworks/broker.rst
+++ b/frameworks/broker.rst
@@ -39,9 +39,6 @@ variable :zeek:see:`Cluster::NodeType`.
 - Worker
 - Proxy
 
-Proxy nodes are optional with Zeek's base scripts. However, certain external or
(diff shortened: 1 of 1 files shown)


ckreibich:
Nice catch. This is pretty bad since nothing complains if you don't spin up proxies — looks like `Cluster::publish_hrw()` [returns `F`](https://github.com/zeek/zeek/blob/master/src/broker/messaging.bif#L229-L230) if there are no nodes in the pool, but we're not actually checking that in any of its uses.

ckreibich (approved):

PR #178, opened 2023-03-22
Explain the new `Site::local_nets` behavior re. private address ranges.

ckreibich:
This expands the section on site-local configuration to cover the relationship of `Site::private_address_space` and `Site::local_nets`, and tweaks bits that suggested edits are required where that is no longer strictly the case.

Complements zeek/zeek#2771 and zeek/zeekctl#54.


quickstart.rst
```diff
@@ -504,32 +504,47 @@ under :samp:`{$PREFIX}/share/zeek/policy/misc/detect-traceroute` and contains a
...
   zeek -r mypackets.trace local -e "Site::local_nets += { 1.2.3.0/24, 5.6.7.0/24 }"
 
-When running with ZeekControl, this value is set by configuring the
-:file:`networks.cfg` file.  Note the example also shows how you can execute
-script code without it being in a :file:`.zeek` file.
+Note how this example executes script code without editing a :file:`.zeek` file,
+via the ``-e`` command-line option.
+
+Zeek provides an exhaustive list of IANA-registered private address space via
+the :zeek:see:`Site::private_address_space` set.  By default Zeek automatically
```
awelzel, line 537: Given we're in `quickstart.rst`, this reads a bit convoluted now and almost misleading currently. On line 522 above, we say:

> For such analysis to work correctly, you need to tell Zeek about your network.

But then down here, one sentence in, we say something like "you're probably fine, because the default is nice and most likely sufficient".


So reshuffeling things probably reads better for an introduction/quickstart. A sketch that comes to mind:

```
zeek -i en0 local
```

Some of Zeek's logic distinguishes networks local to your site from ones elsewhere.  By default, Zeek considers IANA-registered private address space as local. That includes the typical local networks 192.168/.0.016... If you want Zeek to consider networks outside of that space as local, you'll need to tell it about it.  You can do this by configuring [ ... ].

And if you're so advanced (tongue in cheek) that you do not want to consider PAS as local, set `Site::private_address_space_is_local=F` and set `Site::local_nets`

> Note how this example executes script code without editing a :file:`.zeek` file, via the ``-e`` command-line option.

Tangent, I think that would be nicer to use the redef-on-command-line feature directly, rather than `-e`, but it also reads a bit like a distraction if there's a focus on the local nets topic

quickstart.rst
```diff
@@ -504,32 +504,47 @@ under :samp:`{$PREFIX}/share/zeek/policy/misc/detect-traceroute` and contains a
...
+considers this space site-local, adding it to :zeek:see:`Site::local_nets` at
+startup. If your network consists only of private subnets such as 10/8 or
+192.168/16, no additional configuration is required.  If you do not want Zeek to
+consider all standard private address space local, set the
+:zeek:see:`Site::private_address_space_is_local` variable to ``F``. If you do so
+and don't populate :zeek:see:`Site::local_nets`, Zeek prints a warning about
+lacking site configuration.
```
awelzel, line 544: > If you do so
and don't populate :zeek:see:`Site::local_nets`, Zeek prints a warning about
lacking site configuration.

That seemed reasonable to point out when it was the default behavior. I'm don't think we need to mention warnings produced when you change away from defaults and then also do not setting Site::local_nets.

quickstart.rst
```diff
@@ -504,32 +504,47 @@ under :samp:`{$PREFIX}/share/zeek/policy/misc/detect-traceroute` and contains a
...
-warning about lacking the :zeek:see:`Site::local_nets` variable being
-configured. You can supply this information in the :file:`local.zeek` file or
-at the command line like this (supply your “local” subnets in place of the
-example subnets):
+Some of Zeek's logic distinguishes networks local to your site from ones
+elsewhere. For such analysis to work correctly, you need to tell Zeek about your
+network. You do this by configuring the :zeek:see:`Site::local_nets` variable,
+a :zeek:type:`set` of :zeek:type:`subnet` ranges.  You can provide this
+information in the :file:`local.zeek` file or at the command line, as follows
+(using your own networks, of course):
```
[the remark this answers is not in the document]
ckreibich: Argh, I just lost a bunch of my responses here, retyping — I've reworded this whole part.

quickstart.rst
ckreibich: I've shortened it.

> would be nicer to use the redef-on-command-line feature directly

Agree, but I think it's undocumented, and not shown in `zeek -h` either, so I steered clear. I know that we're using it elsewhere in the docs, which isn't that great. The quickstart could use a section "Invoking Zeek" that covers this.

PR #195, opened 2023-06-14
Refresh Spicy analyzer docs for zeek-6.

bbannier:
We have since moved spicy-plugin into Zeek, and extended the zkg packet template.

### DIFF
--- a/devel/spicy-analyzers.rst
+++ b/devel/spicy-analyzers.rst
@@ -3,10 +3,9 @@ Writing analyzers with Spicy
 ============================
 
 `Spicy <https://docs.zeek.org/projects/spicy/en/latest/index.html>`_ is a
-parser generator that makes it easy to create robust C++ parsers network
-protocols, file formats, and more. Spicy analyzers can be integrated with Zeek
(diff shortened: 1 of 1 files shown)


devel/spicy-analyzers.rst
```diff
@@ -4,29 +4,26 @@ Writing analyzers with Spicy
 
 `Spicy <https://docs.zeek.org/projects/spicy/en/latest/index.html>`_ is a
 parser generator that makes it easy to create robust C++ parsers network
```
awelzel, line 6: Unrelated, but think this is missing a "for" after "parsers".

> to create robust C++ parsers "for" network protocols, file formats, ...

awelzel (approved):

timwoj:
Does this need to go into master or just the docs for 6.0?

bbannier:
> Does this need to go into master or just the docs for 6.0?

This definitely affects zeek-6.0.0, so for versioned docs it should go into both `master` and `release/6.0`.

rsmmr (approved):

PR #200, opened 2023-07-13
Update binary package details to reflect new version labeling

ckreibich:
This covers the change in OBS package naming to no longer provide packages abstractly named "zeek" and "zeek-lts", and instead shift to ones named after release trains such as "zeek-6.0".

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -20,8 +20,10 @@ Installing Zeek
 ===============
 
 To run Zeek, grab our official Docker images, download our Linux binary
-packages, install via Homebrew on your Mac, use the ports collections on
-FreeBSD and OpenBSD, or build Zeek yourself.
+packages, install via Homebrew on your Mac, use the ports collections on FreeBSD
+and OpenBSD, or build Zeek yourself. For details about our release cadence
+and the significance of Zeek's version numbers, please refer to our `Release
+Cadence <https://github.com/zeek/zeek/wiki/Release-Cadence>`_ wiki page.
 
 Docker Images
 =============
@@ -30,7 +32,8 @@ We provide official Docker images on Docker Hub at https://hub.docker.com/u/zeek
 
     * For the latest feature release: ``docker pull zeek/zeek:latest``
     * For the latest LTS release: ``docker pull zeek/zeek:lts``
-    * For a specific release: ``docker pull zeek/zeek:5.0.0-rc1``
+    * For the latest release in a given series: ``docker pull zeek/zeek:5.2``
+    * For a specific release: ``docker pull zeek/zeek:6.0.0``
     * For the nightly build: ``docker pull zeek/zeek-dev:latest``
 
 Additionally, we push these images to Amazon's Public Elastic Container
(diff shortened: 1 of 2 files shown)


ckreibich:
I've made corresponding changes over in [the wiki](https://github.com/zeek/zeek/wiki/Binary-Packages), feel free to review/edit as well. I'm not sure we require that wiki page still, thoughts welcome.

install.rst
```diff
@@ -26,11 +26,12 @@ FreeBSD and OpenBSD, or build Zeek yourself.
 Docker Images
 =============
 
-We provide official Docker images on Docker Hub at https://hub.docker.com/u/zeek
+We provide official Docker images on Docker Hub at https://hub.docker.com/u/zeek.
 
     * For the latest feature release: ``docker pull zeek/zeek:latest``
     * For the latest LTS release: ``docker pull zeek/zeek:lts``
```
awelzel, line 32: Should we remove `lts` from here even if we're still pushing it?

install.rst
```diff
@@ -26,11 +26,12 @@ FreeBSD and OpenBSD, or build Zeek yourself.
...
-    * For a specific release: ``docker pull zeek/zeek:5.0.0-rc1``
```
awelzel, line 33: We don't push `-rc1` - I sometimes wished we did though. Worth it?

install.rst
```diff
@@ -92,6 +92,13 @@ on which version you’re using), and includes a complete Zeek environment with
 See our `Binary Packages wiki page <https://github.com/zeek/zeek/wiki/Binary-Packages>`_
 for the latest updates on binary releases.
 
+.. note:: In the past we also provided packages always referring to the latest
+   feature and LTS releases (``zeek`` and ``zeek-lts``, respectively). These
```
awelzel, line 96: This still holds for the feature release IIUC, otherwise how would you install Zeek 6.1 - that continues to be the `zeek` package and maybe should stay on line 75 above.

install.rst
ckreibich: I wouldn't, seems potentially useful for others to build on in their own CI pipelines, for ad-hoc testing, etc. I think this is a little different from the binary images, which are more likely to end up in production use than these images. We could add a note similar to the one for binary images if you're that uneasy about it.

install.rst
ckreibich: Yeah, agree.

install.rst
ckreibich: Perhaps I'm confused — I thought when 6.1 comes out, we'll push `zeek-6.1` so you can say e.g. `apt install zeek-6.1` for the latest in that line of releases. I thought both `zeek` and `zeek-lts` won't be pushed any longer, since they both have the problem of switching release lines at a time of our choosing.

ckreibich:
Btw ... the wiki page stated that when switching e.g. from `zeek` to `zeek-lts`, the currently existing packages would automatically uninstall. That is not the case for the current 5.0 and 6.0 ones, at least not when I tested with Ubuntu, so I removed that comment. The packages currently conflict if you install one line over the other. It'd be nice if it could work that way though, or (better?) if they could install alongside each other so the alternatives system can front them.

@0xxon, fyi.

install.rst
awelzel: Hmm, my statement was based on @0xxon 's comment [elsewhere](https://zeekorg.slack.com/archives/GUCJ5GH5M/p1688973606900199?thread_ts=1688712362.274209&cid=GUCJ5GH5M):
> Johanna
> Well, I am happy to make a zeek-5.0 then

> Johanna
> I don’t want to do packages for the feature releases though :slightly_smiling_face:

> Johanna
> that would get cluttered quick


So my understanding of is different.  Latest feature release (and also x.0 when overlapping) will continue to be available as `zeek` package only. The new `zeek-x.0` packages are only available for LTS releases.

0xxon: Yup, my understanding was the one of Arne. I am happy to push `zeek-x.0` packages for the LTS releases - but I want to avoid pushing our feature releases too, for several reasons.

So my current plan would be to have a `zeek` package for the current feature release (and people that just want to update automatically), and `zeek-x.0` packages for people who want to install the last LTS release.

0xxon:
Yeah, the wiki is wrong there, I will update that.

We can move towards using alternatives - I admittedly never thought about that option. That will be some work though - and will probably not happen in the super short term, unless you think this should be a priority.

0xxon:
Also - how long do we want to keep the `-lts` package around? We should probably give people _plenty_ of warning - I assume that there might be people using it. My gut feeling is to wait till at least Zeek 7.0 to retire it.

install.rst
awelzel: > So my current plan

I actually like that plan a lot. Any `x.1` or `x.2` feature release goes out of maintenance on the next release immediately, so the choice for users should either be:
* Run the latest feature release via the `zeek` package
* Run the latest LTS release with explicit version, the `zeek-x.0` package. Preferably the latest one, but prior ones will be around.
 
...and prior ones might even be supported for the short period where Zeek 5.0 and Zeek 6.0 (currently) are both receiving updates still. Currently we do not have a good packaging story for that, but with zeek-6.0 and zeek-7.0 existing there will be.

ckreibich: I see, thanks. I like this as a forcing function toward the LTS releases, and as a signal that x.1/x.2 are short-lived.

But I also think our limited support windows for x.1/x.2 and pulling people across release lines are separate concerns, and ideally we'd address both needs ("stick to x.y, even if I fall out of maintenance", which favors stability, vs "just give me latest", which favors development). Then again it's reasonable for us to argue that folks wanting stability are better off with "pinned" LTS lines overall, plus installations that rarely update (to whatever version, including latest x.y) will generally need to look closely at what it means to upgrade when they do, anyway.

On a related note, if we were to start publishing migration guides, I'd be strongly in favor of doing those only for LTS releases, so your plan aligns with that.

So — I remain a bit torn, but I'm fine with it. I'll update the wording.

ckreibich:
> My gut feeling is to wait till at least Zeek 7.0 to retire it.

Yup, I agree.

ckreibich:
I finally updated this — sorry for the delay. I also addressed a Dependabot bump for GitPython while I was at it.

PR #206, opened 2023-08-24
conf: Add sphinx_rtd_theme to extensions, remove env based import

awelzel:
The RTD build failed with "NameError: name 'html_theme' is not defined".

Remove the conditional import and align to what Spicy and the sphinx_rtd_theme docs [1] indicate should be done.

[1] https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html

RTD build: https://readthedocs.org/projects/zeek-docs/builds/21704944/

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -141,15 +141,7 @@ highlight_language = 'none'
 
 # -- Options for HTML output ---------------------------------------------------
 
-# The theme to use for HTML and HTML Help pages.  See the documentation for
-# a list of builtin themes.
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
-
-if not on_rtd:
-    # only import and set the theme if we're building docs locally
-    import sphinx_rtd_theme
-    html_theme = 'sphinx_rtd_theme'
-    html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+html_theme = 'sphinx_rtd_theme'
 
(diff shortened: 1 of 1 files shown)


awelzel:
Something must have changed on the RTD side - I've ended up here which is from a few days ago:
https://github.com/readthedocs/readthedocs.org/pull/10638

The build above succeeded, so I think this still makes sense.

conf.py
```diff
@@ -28,6 +28,7 @@
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions += ['zeek', 'sphinx.ext.todo', 'zeek_pygments', 'literal-emph']
 extensions += ['sphinx.ext.extlinks']
+extensions += ['sphinx_rtd_theme']
```
bbannier, line 31: Just for my own curiosity, we don't seem to have this over in Spicy and the build still succeeded (just triggered a test build). Why are we adding this?

bbannier (approved):

conf.py
awelzel: It's shown in the theme docs: https://sphinx-rtd-theme.readthedocs.io/en/latest/installing.html

But then there is this:

> Adding this theme as an extension is what enables localization of theme strings in your translated output. If these strings are not translated in your output, either we lack the localized strings for your locale, or you are using an old version of the theme.

So, let me drop it given no localization here. Thanks!

PR #207, opened 2023-08-24
Clarifying differing semantics of vector += if RHS is a compatible vector

vpax:
As mentioned in the discussion of https://github.com/zeek/zeek/issues/3191, the Zeek documentation is incomplete in its discussion of using the `+=` operator when the RHS is a vector compatible with the LHS. This tweak aims to remedy that.


awelzel (approved):

awelzel:
Squashed the fixup before merge.

PR #211, opened 2023-09-19
Import Spicy's documentation on its Zeek integration.

rsmmr:
These docs will live on the Zeek-side going forward, along with the corresponding code that's now part of Zeek already.

Closes #3291.


devel/spicy-analyzers.rst
```diff
@@ -1,35 +0,0 @@
...
-
-Spicy is documented separately, so this section just presents high-level points
-relevant for Zeek.
-
-Installation
-============
-
-A Zeek configured with default ``./configure`` options includes Spicy.
-
-Writing an analyzer
```
bbannier, line 18: I think this section (especially its first paragraph) was valuable since it gave a quick and high-level introduction. The documentation in this patch instead builds all this up in a very systematic way and only at the very end shows how to actually create a package.

A lot of effort went into the package templates for Spicy analyzers so that it probably is a pretty good starting point. Do you think it would make sense to keep this paragraph as a TLDR; at the top of the Zeek Spicy docs?

devel/spicy/autogen-spicy-docs
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
bbannier, line 1: This file seems unused.

conf.py
```diff
@@ -26,7 +26,7 @@
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions += ['zeek', 'sphinx.ext.todo', 'zeek_pygments', 'literal-emph']
+extensions += ['zeek', 'sphinx.ext.todo', 'zeek_pygments', 'spicy-pygments', 'literal-emph']
```
bbannier, line 29: The `spicy-pygments` extension currently prevents parallel builds. I opened zeek/spicy#1530 to fix that.

bbannier (approved):

devel/spicy/autogen-spicy-docs
rsmmr: This must be called manually as it needs access to files outside of the repository (similar to how it was inside the Spicy repo).

devel/spicy-analyzers.rst
rsmmr: Makes sense, I've moved that back in.

devel/spicy/autogen-spicy-docs
bbannier: Currently this scripts seems to be mentioned anywhere which is not great for making sure that the docs stay up to date. I am particular worried about changes landing in e.g., spicy-tftp and the Zeek docs becoming stale.

Over in zeek/spicy we [automated checking out the needed repos when building the docs](https://github.com/zeek/spicy/blob/30fb6309338b9bedd329c35c593fbad516748038/doc/Makefile#L64-L72), could we add something similar here? 

If that is not possible, the script should at least be called out prominently in `README`.

rsmmr: Ok, I added that to the docs GH action. Not sure if that's the best place, but doing it Zeek-side isn't feeling great either.

rsmmr: Well, never mind, that doesn't work because the script needs `spicy-doc` built. I'll leave the Makefile target in but remove from CI; and check if I can trigger it Zeek-side after all.

PR #215, opened 2023-10-11
Add developer documentation on how to reformat long-lived forks

bbannier:


bbannier:
I am definitely open for suggestions on where to place this. For tracking and reviewing I did want to add this to the repo, but we could move it to the wiki as well.

devel/contributors.rst
```diff
@@ -31,3 +31,83 @@ Checking for Memory Errors and Leaks
...
+   git checkout master -- .pre-commit-config.yaml .clang-format .style.yapf .cmake-format.json
+   git commit -m 'Bump formatter configurations'
+
+   # With a fork branched from upstream at commit FORK_COMMIT, rebase the
+   # config update commit 'Bump formatter configurations' to the start of the
+   # fork, but do not yet rebase on master (interactively move the last patch
+   # to the start of the list of patches).
+   git rebase -i FORK_COMMIT
+
+   # Reformat all commits accoring to configs at the base. We use the `--exec`
```
timwoj, line 100:
```suggestion
   # Reformat all commits according to configs at the base. We use the `--exec`
```

timwoj (approved):
I think where you have it is fine. There is other existing documentation in the same place, so it makes sense to include this here too.

awelzel (approved):
I think this is great (and useful information in general).

PR #222, opened 2023-11-03
Fix Github edit links in top-right of HTML pages

ckreibich:
Matti noticed that these links 404'd because they omitted the file suffix. This change uses a different variable name in Jinja to fix that, and also changes the target link to the editable sources instead the default source viewer.

There's an LTD test build here if you want to check it out: https://docs.zeek.org/en/topic-christian-test-fix-github-edit-link/

Once this goes in we need it twice more, for the Broker and zkg docs.


matti:
yeah its annoying

PR #223, opened 2023-11-06
Add log page to document ldap.log and ldap_search.log

mbispham:
PR to add ldap.log and ldap_search.log documentation to RTD logs dir.

### DIFF
--- a/logs/ldap.rst
+++ b/logs/ldap.rst
@@ -22,8 +22,7 @@ LDAP communicates using a client-server model. The LDAP server contains the dire
 
 The LDAP analyzer outputs two LDAP related logs. :file:`ldap.log` contains details about the LDAP session except those related to searches. :file:`ldap_search.log` contains information related to LDAP searches.
 
-For details on every element of the :file:`ldap.log` and :file:`ldap_search.log` refer to :zeek:see:`LDAP`.
-:file:`ldap.log`. Below is an inspection of the :file:`ldap.log` and :file:`ldap_search.log` in JSON format.
+For details on every element of the :file:`ldap.log` and :file:`ldap_search.log` refer to :zeek:see:`LDAP`. Below is an inspection of the :file:`ldap.log` and :file:`ldap_search.log` in JSON format.
 
 ldap.log
 ========
--- a/logs/ldap.rst
+++ b/logs/ldap.rst
@@ -1,0 +1,91 @@
+============================
+ldap.log and ldap_search.log
+============================
+
+The Lightweight Directory Access Protocol (LDAP) log, or :file:`ldap.log`, is a widely observed protocol
+commonly used for authenticating, directory lookups, centralizing organisational information and accessing client information on email servers. Accordingly, the protocol attracts significant attention from those with adversarial intention.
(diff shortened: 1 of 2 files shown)


logs/ldap.rst
```diff
@@ -0,0 +1,90 @@
...
+      "id.orig_h": "10.0.0.1",
+      "id.orig_p": 25936,
+      "id.resp_h": "10.0.0.2",
+      "id.resp_p": 3268,
+      "message_id": 3,
+      "version": 3,
+      "opcodes": ["bind simple"],
+      "results": ["success"],
+      "objects": ["CN=xxxxxxxx,OU=Users,OU=Accounts,DC=xx,DC=xxx,DC=xxxxx,DC=net"],
+      "arguments": ["REDACTED"]
```
awelzel, line 51: @mbispham - do you have insight (or access to production data logs) whether using sets/vectors for these fields is reasonable? Could these be scalar values instead. If there's a few special cases here and and there, we could consider unrolling them instead.

Main thought is ingestion/filtering of multi-value fields that may be annoying for downstream consumers when in practice there's only a single value (same for ldap_search.log).

Thanks for the PR! Much appreciated!

(cc @mmguero)

bbannier: This [aggregates state across multiple messages](https://github.com/zeek/zeek/blob/4eb1b71d1b29845dec216536d830e79f2f95959b/scripts/base/protocols/ldap/main.zeek#L107-L110), so I think unless we want to new log syntax this should be a container.

These fields already contain [string representations of more complicated LDAP constructs](https://github.com/zeek/zeek/blob/4eb1b71d1b29845dec216536d830e79f2f95959b/testing/btest/Baseline/scripts.base.protocols.ldap.attributes/ldap_search.log#L10) which I believe were modeled after what Wireshark does. We should check whether Wireshark does anything interesting for lists (though I doubt that since they often do not aggregate such state across packages), but users seem to expect Wireshark format, https://github.com/zeek/spicy-ldap/issues/21, https://github.com/zeek/spicy-ldap/issues/23, https://github.com/zeek/spicy-ldap/issues/27.

awelzel: > This [aggregates state across multiple messages](https://github.com/zeek/zeek/blob/4eb1b71d1b29845dec216536d830e79f2f95959b/scripts/base/protocols/ldap/main.zeek#L107-L110), so I think unless we want to new log syntax this should be a container.

That table is used to correlate requests and responses based on message_id. A server may reply to requests out-of-order so that makes sense.

Using vectors for opcode, result and all the other fields is a different question: If a request/response pair is only ever about a single opcode (or an error), then these shouldn't be modeled as vectors. With the exception of "search" (which is treated separately) the RFC indicates this is the case.

I'm mostly just sensitive as it's very related to the tx_hosts / rx_hosts fields that were removed from the files.log: In practice, they only ever contained a single entry, but being represented as arrays they needed special treatment downstream.

mbispham: @awelzel - I think this is a worthwhile discussion - but not for this PR.

PR #228, opened 2023-11-20
management/full-config: no toml highlighting

awelzel:
With Pygments 2.17, the TOML parser was rewritten[1, 2]. It now fails to parse and highlight the `full-config.ini` file. The key-only `agent-testbox` within `[instances]` makes the standard toml parsing barf, too. Flip to ini.

[1] https://pygments.org/docs/changelog/#version-2-17-0
[2] https://github.com/pygments/pygments/pull/2576

### DIFF
--- a/frameworks/management.rst
+++ b/frameworks/management.rst
@@ -606,11 +606,7 @@ Here's a full-featured configuration describing the available options, assuming
 a single agent running on a machine "testbox" with default settings:
 
 .. literalinclude:: management/full-config.ini
-   :language: toml
-..
-   Using toml in the above only because it produces slightly better highlighting.
-   The files are not technically TOML but what's compatible with Python's
-   configparser classes.
+   :language: ini
 
 .. _simplification-instance-local:
 


bbannier:
Just wondering whether I am reading the [Wikipedia article on the INI format](https://en.wikipedia.org/wiki/INI_file) correctly, INI does not usually have value-less entries, right?

Regardless, the file definitely isn't TOML (e.g., strings are not quoted).

bbannier (approved):

bbannier:
Left a :shipit:, but per my above comment, I think it would be great to use an INI format with less extensions (i.e., here: always require values). Could you put an issue for that onto our pile @awelzel?

timwoj:
> Regardless, the file definitely isn't TOML (e.g., strings are not quoted).

FWIW, our configuration files have never been completely TOML-compliant. They're TOML-ish, but not really TOML.

ckreibich:
Hey thanks for fixing this — what a headache. Per my former comment in the code I was really just looking for something that does a decent job highlighting the file. It was always simple Python convenience that wins out here. If there's any doubt here what to use, then let's simply not highlight at all.

> FWIW, our configuration files have never been completely TOML-compliant. They're TOML-ish, but not really TOML.

Yeah.

PR #230, opened 2023-11-23
Reduce some details in the `table[pattern]` usage note

ckreibich:
This is quite nit-picky, forgive me — but it tones down the details a bit because I don't think it's clear to most users how one would actually construct an adversarial set of inputs to a given table. (We could explain that too, though I'm thinking that's not really necessary here.)


awelzel:
Thanks!

PR #232, opened 2023-11-24
Adding a QUIC section to ssl.rst

mbispham:
Added a QUIC section. Trying to emphasize the presence of a server name in TLS 1.3.


logs/ssl.rst
```diff
@@ -358,6 +358,48 @@ practitioners concerned by the loss of visibility, and the opportunity for
...
+
+QUIC
+====
+
+.. versionadded:: 6.1
+
+The QUIC protocol integrates encryption, stream multiplexing and flow control
+at the transport layer. QUIC uses TLS 1.3 by default. The QUIC analyzer outputs
+logs to :file:`ssl.log`. The QUIC analyzer attempts to handle the INITIAL packet
+payload to provide greater observability.
```
awelzel, line 370: Thanks for the PR - I'd think a small `logs/quic.rst` file, noting that `ssl.log` entries are created as well would better fitting.

There's a separate `quic.log`. Anything showing in `ssl.log` is because the QUIC analyzer forwards data to the SSL analyzer, not because it explicitly writes to the `ssl.log`.

mbispham:
This needs some squashing and line wrapping...I wanted to check my understanding of the SCID field. In [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#name-new_connection_id-frames) the ID fields can change to adapt to new conditions but that the analyzer will continue to use the same SCID seen in the initial packet. Is that correct?

awelzel:
> This needs some squashing and line wrapping...I wanted to check my understanding of the SCID field. In [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#name-new_connection_id-frames) the ID fields can change to adapt to new conditions but the analyzer will continue to use the same SCID seen in the initial packet. Is that correct?

The analyzer should send the actual SCID for a given packet to the script layer. On the script side we simply log the latest one seen for a given connection. Reality and corner-cases likely aren't as straightforward. If you have pcap/scenario and an idea how the behavior should be (maybe a new log entry when the SCID), this could certainly be changed. For RETRY packets, a new log entry should already be created here, even if the connection is the same.

https://github.com/zeek/zeek/blob/0b5126f6507fc7af1a6f5ec23587d4246a8b970b/scripts/base/protocols/quic/main.zeek#L125-L126

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
+====
+QUIC
+====
+
+.. versionadded:: 6.1
+
+The QUIC protocol integrates encryption, stream multiplexing and flow control at
+the transport layer. QUIC uses TLS 1.3 by default. The QUIC spicy analyzer
```
awelzel, line 8: No need to mention spicy.
```suggestion
the transport layer. QUIC uses TLS 1.3 by default. Zeek's QUIC analyzer
```

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+attempts to handle the QUIC INITIAL packet payload to provide greater
+observability into the protocol, including the server name.
```
awelzel, line 10: maybe shorten as "to provide visibility into the TLS handhake".

It does a few things more like retry packet handling, but yes, it's not a full protocol analyzer.

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+
+
+- **client_initial_dcid**: When QUIC initiates a connection it uses Random
+  Number Generators to create the first Destination Connection ID (DCID). This
+  DCID is subsequently used for routing and packet protection by client and
+  server.
+
+
+- **server_scid**: A QUIC supported server responds to a DCID by
+  selecting a Source Connection ID (SCID). This usually occur within the
```
awelzel, line 58:
```suggestion
  selecting a Source Connection ID (SCID). This usually occurs within the
```

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+  server’s first INITIAL packet. This is typically used by the client in
+  subsequent packets, although the SCID can change to adapt to new network
+  conditions.
+
+
+- **client_protocol**: During the Application-Layer Protocol Negotiation
+  (ALPN) extension, if the ClientHello packet is available, it is extracted.
+  In the example above, we can see that HTTPv3 is the client protocol.
```
awelzel, line 66: This is a bit backwards. I'd say "If the ClientHello packet is successfully extracted and contains the ALPN extension, the extension's first entry is placed in `client_protocol`.

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+
+
+- **history**: An experimental feature provides a history of the QUIC
+  protocol usage. In the example above the history outlines:
+
+    + An initial packet from the client - a new connection
+
+
+    + An SSL Client/Server Hello from the client - the start of a
```
awelzel, line 75:
```suggestion
    + An SSL ClientHello from the client (S) - the start of a
```


etc.

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+
+- **client_protocol**: During the Application-Layer Protocol Negotiation
+  (ALPN) extension, if the ClientHello packet is available, it is extracted.
+  In the example above, we can see that HTTPv3 is the client protocol.
+
+
+- **history**: An experimental feature provides a history of the QUIC
+  protocol usage. In the example above the history outlines:
+
+    + An initial packet from the client - a new connection
```
awelzel, line 72:
```suggestion
    + An initial packet from the client (I) - a new connection
```

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    + An initial packet response from the server - an acknowledgement
+      from the server of the new connection
+
+
+    + An SSL Client/Server Hello response from the server - the
+      selection  of a cipher suite from the options provided by the
+      client
+
+    + A handshake packet - completes the TLS negotiation
```
awelzel, line 87:
```suggestion
    + A handshake packet from the client (H)
```

I don't think these necessarily complete the negotiation, so would leave it out. Also might be good to avoid mixing SSL and TLS and stick with one of them in the docs.

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    + An SSL Client/Server Hello from the client - the start of a
+      SSL/TLS handshake
+
+
+    + An initial packet response from the server - an acknowledgement
+      from the server of the new connection
+
+
+    + An SSL Client/Server Hello response from the server - the
```
awelzel, line 83:
```suggestion
    + An SSL ServerHello response from the server (s) - the
```

logs/quic.rst
mbispham: ✅

logs/quic.rst
mbispham: ✅

logs/quic.rst
mbispham: ✅

logs/quic.rst
mbispham: ✅

logs/ssl.rst
```diff
@@ -358,6 +358,49 @@ practitioners concerned by the loss of visibility, and the opportunity for
...
+    "ssl_history": "Cs"
+  }
+
+The QUIC analyzer is an initial attempt to provide observability into QUIC
+network traffic. In Zeek 6.1, although included by default, the QUIC analyzer is
+still considered experimental. In addition to code contributions, feedback and
+issues are greatly appreciated, and can be reported via Zeek github, slack or
+discourse - https://zeek.org/community/.
+
+
```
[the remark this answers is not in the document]
mbispham: Yeah - it's gone. One to ponder is whether `ssl.log` has a field identifying the underlying transport protocol.

logs/quic.rst
```diff
@@ -0,0 +1,99 @@
...
+
+
+
+Conclusion
+==========
+
+The QUIC analyzer is an initial attempt to provide observability into QUIC
+network traffic. In Zeek 6.1, although included by default, the QUIC analyzer is
+still considered experimental. In addition to code contributions, feedback and
+issues are greatly appreciated, and can be reported via Zeek github, slack or
```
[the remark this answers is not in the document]
mbispham: ✅

logs/quic.rst
mbispham: ✅

logs/quic.rst
mbispham: ✅

logs/quic.rst
mbispham: ✅

PR #245, opened 2024-01-31
Point to openssl 3.1.1 in Windows instructions

philrz:
I was rebuilding Zeek on Windows to reproduce a problem and ended up bumping into the same problem described in https://github.com/zeek/zeek/pull/3587. I propose updating the user-facing docs so others don't get similarly thrown by this.


timwoj:
Fixed separately in 3a27cba0e42939ab074ab68705402d8a6a2bdba4

PR #250, opened 2024-02-27
types/statements: Document delete of all elements

awelzel:


script-reference/types.rst
```diff
@@ -1424,6 +1441,11 @@ false.
 See the :zeek:keyword:`for` statement for info on how to iterate over
 the elements in a vector.
 
+.. versionadded:: 7.0
+
+See the :zeek:keyword:`delete` statement can be used to delete all elements
```
timwoj, line 1446:
```suggestion
See the :zeek:keyword:`delete` statement, which can be used to delete all elements
```

timwoj (approved):

script-reference/types.rst
awelzel: I'm going for "The delete statement can be used to...", I never meant to copy the "See the" from above..

PR #252, opened 2024-03-11
Expand documentation on exporting Spicy types

bbannier:


devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
 parsing. Unit variables are non-optional by default, unless declared
 as ``&optional`` in Spicy.
 
+.. _spicy_unit_export:
+
+Controlling created Zeek types
+""""""""""""""""""""""""""""""
+
+If needed the automatic unit type exporting described above can be customized. The general syntax for this is
```
rsmmr, line 610:
```suggestion
If needed the automatic unit type exporting described above can be customized. The general syntax for this is:
```

devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
...
+
+.. code-block:: spicy-evt
+
+    export SPICY_ID [with { [SIPCY_FIELD_NAME [&log], ]... }];
+    export SPICY_ID [as ZEEK_ID (with { [SIPCY_FIELD_NAME [&log], ]...) }];
+
+.. rubric:: Controlling exported fields
+
+Assume we are given a Spicy unit type ``foo::X``
```
rsmmr, line 619:
```suggestion
Assume we are given a Spicy unit type ``foo::X``:
```

devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
...
+
+    module foo;
+
+    public type X = unit {
+        x: uint8;
+        y: uint8;
+        z: uint8;
+    };
+
+To create a Zeek type with only the field ``x`` to following ``export`` statements are equivalent:
```
rsmmr, line 631:
```suggestion
To create a Zeek type without the field ``x`` the following ``export`` statements are equivalent:
```

rsmmr: Higher level, I find it confusing to group all these 4 cases together because they all have different purposes, whereas the "are equivalent" seems to suggest they are all doing the same and it doesn't matter which one to use. Suggestion: walk through the 4 cases invididually: (1) skip a field, (2) include only selected fields, (3) rename the record.

devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
...
+    export foo::X with { y, z };
+    export foo::X as foo::X with { y, z };
+    export foo::X without { x };
+    export foo::X as foo::X without { x };
+
+.. rubric:: Adding Zeek ``&log`` attributes to created Zeek types
+
+Data intended to be written to Zeek log streams needs to be marked ``&log``. We
+can trigger creating of this attribute in exported types either per-field or
+for the whole ``record``.
```
rsmmr, line 644:
```suggestion
In Zeek records, data intended to be written to Zeek log streams needs to be marked ``&log``. We
can trigger creation of this attribute from Spicy either per-field or
for the whole ``record``.
```

devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
...
+
+The forms
+
+.. code-block:: spicy-evt
+
+    export foo::X &log;
+    export foo::X as foo::X &log;
```
rsmmr, line 651:
```suggestion
    export foo::X as foo::X &log;
```

Similar to above, why include the `as ...` case here?

devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
...
+
+      type X: record {
+        x: count &optional &log;
+        y: count &optional &log;
+        z: count &optional &log;
+      };
+
+    }
+
+To mark individual fields ``&log`` we would use the attribute on the respective fields, e.g.,
```
rsmmr, line 669:
```suggestion
To mark individual fields ``&log`` we can use the attribute on the respective fields, e.g.,
```

devel/spicy/reference.rst
```diff
@@ -602,6 +602,91 @@ optional on the Zeek-side, as they may not have been set during
...
+
+Controlling created Zeek types
+""""""""""""""""""""""""""""""
+
+If needed the automatic unit type exporting described above can be customized. The general syntax for this is
+
+.. code-block:: spicy-evt
+
+    export SPICY_ID [with { [SIPCY_FIELD_NAME [&log], ]... }];
+    export SPICY_ID [as ZEEK_ID (with { [SIPCY_FIELD_NAME [&log], ]...) }];
```
rsmmr, line 615:
```suggestion
    export SPICY_ID [as ZEEK_ID] [with { [SPICY_FIELD_NAME [&log], ]... }];
```

devel/spicy/reference.rst
bbannier: Good idea. I have completed the grammar and added separate explanations of the different uses cases. With that there is no need to show the "equivalent" forms below. I kept the verbose use cases below, but now mark them _Example_.

devel/spicy/reference.rst
```diff
@@ -598,10 +598,118 @@ dependent types must be exported *first* in the EVT file for this to
...
+The general syntax for this is:
+
+.. code-block:: spicy-evt
+
+    export SPICY_ID [with { [SIPCY_FIELD_NAME [&log], ]... }];
+    export SPICY_ID [as ZEEK_ID [ with { [SIPCY_FIELD_NAME [&log], ]...]  }];
+    export SPICY_ID [without { [SIPCY_FIELD_NAME, ]... }];
+    export SPICY_ID [as ZEEK_ID [ without { [SIPCY_FIELD_NAME, ]...]  }];
+    export SPICY_ID &log;
+    export SPICY_ID as ZEEK_ID &log;
```
rsmmr, line 620: I'd still try to compress this into a single line but may be matter of taste.

rsmmr (approved):
Looks good now!

devel/spicy/reference.rst
bbannier: That's what I initially tried, but it always overflowed the box in the rendered pages so it required horizontal scrolling, and also required a lot of nesting. I was unsure whether adding line breaks for readability would be confusing to readers. The full expression was

```
export SPICY_ID [&log | with { [SPICY_FIELD_NAME [&log], ]...} | without { [SPICY_FIELD_NAME, ]... } | as ZEEK_ID [&log | { [SPICY_FIELD_NAME [&log], ]...} | without { [SPICY_FIELD_NAME, ]...}]]
```

PR #261, opened 2024-05-06
Regenerate Spicy docs.

rsmmr:
This goes with https://github.com/zeek/zeek/pull/3719.


timwoj:
The macOS failure in CI is fixed in master now.

PR #264, opened 2024-06-19
Document new Spicy API for accessing Zeek-side globals.

rsmmr:


devel/spicy/reference.rst
```diff
@@ -841,6 +841,102 @@ processing:
...
+that could end up producing a very tight coupling of Spicy and Zeek
+code. Use events to communicate from Spicy to Zeek instead.
+
+Access to a Zeek global can be relatively slow as it performs the ID
+lookup and data conversion. Accordingly, the mechanism is primarily
+meant for configuration-style data, not for transferring heaps of
+dynamic state. However, as a way to speed up repeated access slightly,
+you can use :ref:`zeek::get_value <spicy_get_value>` to cache the
+result of the ID lookup, then use ``as_<TYPE>()`` on the cached value
+to retrieve the actual value.
```
bbannier, line 938: I wonder whether we should more explicitly call out a pattern on how to use Zeek values in parsers (e.g., with an dedicated example right here). A good pattern might be to capture a Zeek `option` in a Spicy `global`, and then reference that global during parsing (an example could be a value for a `&max-size` attribute). This trades repeated, potentially costly calls into Zeek with a less costly access to a `global`.

bbannier (approved):
Looks good modulo the doc suggestions I made on zeek/zeek#3788 which would end up here via autogenerated docs.

devel/spicy/reference.rst
rsmmr: It's a bit tricky, I'm not sure I can tell yet what a "good pattern" is. For example, caching an `option` in a global might *not* be the best way to go: options are designed to potentially change at runtime (vs consts, which don't). I think I'd prefer to leave this open for now and see how people end up using it.

rsmmr: Actually, let me add couple of sentences along those lines.

PR #267, opened 2024-07-26
Add a chapter on script optimization

ckreibich:
The goal here is to make regular users aware of the feature at this time, but leave most of the tunables and flags out of the picture for the time being.

@vpax fyi — I'm sure you'll want to iterate on this, or at least review. It's heavily based on your in-tree README, as discussed. (No rush, it can wait until you're back.)

### DIFF
--- a/scripting/optimization.rst
+++ b/scripting/optimization.rst
@@ -1,0 +1,89 @@
+.. _zam:
+
+===================
+Script Optimization
+===================
+
+.. versionadded:: 7.0
+
+.. note::
+
+   ZAM has been available in Zeek for a number of releases, but as of Zeek 7
+   it has matured to a point where we encourage regular users to explore it.
+
+Introduction
+============
+
+The `Zeek Abstract Machine`, "ZAM", is an optional script optimization engine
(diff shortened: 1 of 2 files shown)


scripting/optimization.rst
```diff
@@ -0,0 +1,89 @@
...
+
+A roughly 13% improvement in runtime.
+
+Other Optimization Features
+===========================
+
+You can tune various features of ZAM via additional options to ``-O``, see the
+output of ``zeek -O help`` for details. For example, you can study the script
+transformations ZAM applies, and use ZAM selectively in certain files (via
+``-0|--optimize-files``) or functions (via ``-o|--optimize-funcs``).  Most users
```
vpax, line 88: Nit: the `-0` and `-o` are just because the `getopt` parser needs single-character options to communicate what's seen (at least, that's my understanding). So they could be removed here, as they're not very mnemonic anyway.

vpax: Looks good, could address the above nit.

ckreibich: Thanks, will do!

PR #268, opened 2024-07-27
Edit pass over telemetry docs

ckreibich:
This mainly expands the framework's documentation to give a bit more context on Prometheus, and moves the description of how to get telemetry out of Zeek up, before we give examples of the framework's use (which I think matters more for developers).

I noticed a few discrepancies while updating the code examples:
- The exposition format is now cleaner. Trailing fractional zeroes are pruned, and we don't include timestamps any more (which are optional in the exposition format's definition, and actually looks preferable to me)
- There's a minor bug in the framework in that `MetricOpts` says the `unit` field is optional, but omitting it actually causes errors in the metrics registration functions (just search for unguarded access to `$unit` — `opts$unit). I think that fix can wait for the first point release, particularly since existing code will still set the unit.

This was a nice reminder that we'd catch such things earlier if we had a setup that automatically runs the documentation's code examples as btests, or some such. 

The PR lso updates the management framework side, though the changes here are much smaller.


timwoj:
> The exposition format is now cleaner. Trailing fractional zeroes are pruned, and we don't include timestamps any more (which are optional in the exposition format's definition, and actually looks preferable to me)

Interesting. Is that a prometheus-cpp thing that strips the trailing zeroes?

timwoj (approved):
Looks good to me. I had a couple of minor nits I'll fix during merge. I'll also cherry-pick this over into the 7.0 release branch.

PR #269, opened 2024-07-31
RTD maintenance

awelzel:


conf.py
```diff
@@ -143,6 +143,15 @@
 
 html_theme = 'sphinx_rtd_theme'
 
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
+
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
```
bbannier, line 150: Nit: the default value for the `default` parameter is `None`, no need to tweak it here.

```suggestion
if os.environ.get("READTHEDOCS") == "True":
```

conf.py
```diff
@@ -143,6 +143,15 @@
 
 html_theme = 'sphinx_rtd_theme'
 
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
```
bbannier, line 147: Do we [actually need to set this](https://docs.readthedocs.io/en/stable/guides/canonical-urls.html#sphinx)? Asking since we do not currently set it for the Spicy docs.

bbannier (approved):

conf.py
awelzel: Hmm.. the email said:

> Are my projects affected?
> You have at least one project with a canonical custom domain defined. After the deprecation date, Read the ? Docs will stop defining [html_baseurl](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl) Sphinx configuration, which means that you will need to define it by yourself to keep the canonical custom domain properly configured.
> 
> We found the following projects where you are a maintainer are impacted by this change:
>
> [zeek-docs](https://readthedocs.org/projects/zeek-docs/)

The docs you linked might be dated?

Spicy is a subproject of Zeek, so maybe there pre-notification hasn't detected that, or it's not needed for subprojects (though that would be confusing, too).


Screenshot from 2024-07-31 12-04-20

conf.py
awelzel: It's from their docs, so didn't tweak it. Will leave as is.

(Wonder if there's some type-checking background to passing `""` as default. (without it's typed as `Optional[str]`, with it's `str`).

PR #272, opened 2024-08-12
spicy: Update docs for port/ports removal from .evt files

awelzel:


awelzel:
This is for the updates at https://github.com/zeek/spicy-tftp/pull/11- I've already autogen'ed the non-merged branch.

devel/spicy/getting-started.rst
```diff
@@ -56,16 +56,14 @@ We do both of these by creating an additional control file for Zeek:
...
     :language: spicy-evt
 
-The first block (lines 1-3) tells Zeek that we have a new protocol
-analyzer to provide. The analyzer's Zeek-side name is
-``spicy::MyHTTP``, and it's meant to run on top of TCP connections
-(line 1). Lines 2-3 then provide Zeek with more specifics: The entry
+The first block (lines 1-2) tells Zeek that we have a new protocol
+analyzer to provide. Line 1 defines the analyzer's Zeek-side name as
+``spicy::MyHTTP`` and that it's meant to run on top of TCP connections.
+Line 2 provide Zeek with more specifics: The entry
```
timwoj, line 62:
```suggestion
Line 2 provides Zeek with more specifics: The entry
```

I would actually just rewrite this line entirely. The second part was removed so it's no longer a list of specifics being provided.

> Line 2 provides Zeek with the entry point...

devel/spicy/getting-started.rst
```diff
@@ -91,6 +91,21 @@ having the expected signature of four parameters matching the types of
...
+.. note::
+
+  .. versionchanged:: 7.1
+
+     In prior Zeek versions, the ``.evt`` file supported the keywords
+     ``port`` and ``ports`` to pass information about analyzer activation
+     to Zeek at ``.hlto`` loading time.
+
+     This functionality was deprecated with Zeek 7.0 and removed with Zeek 7.1
+     in favor of using explicit Zeek script side analyzer registration.
```
timwoj, line 107:
```suggestion
     in favor of using explicit Zeek script-side analyzer registration.
```

devel/spicy/tutorial.rst
```diff
@@ -319,7 +328,8 @@ Zeek Script
 -----------
 
 Analyzers normally come along with a Zeek-side script that implements
-a set of standard base functionality, such as recording activity into
+a set of standard base functionality, such as registering the analyzer
+for well known ports and recording activity into
```
timwoj, line 332:
```suggestion
for well-known ports and recording activity into
```

devel/spicy/tutorial.rst
```diff
@@ -82,34 +82,43 @@ protocol analyzer. We follow the same scheme here and put the
...
 application analyzer on top of UDP (``over UDP``). ``TFTP::Packet``
 provides the top-level entry point for parsing both sides of a TFTP
-connection. Furthermore, we want Zeek to automatically activate our
-analyzer for all sessions on UDP port 69 (i.e., TFTP's well known
-port). See :ref:`spicy_evt_analyzer_setup` for more details on defining
+connection. See :ref:`spicy_evt_analyzer_setup` for more details on defining
 such a ``protocol analyzer`` section.
 
+Furthermore, we want Zeek to automatically activate our analyzer for all
+sessions on UDP port 69 (i.e., TFTP's well known port) using the following
```
timwoj, line 96:
```suggestion
sessions on UDP port 69 (i.e., TFTP's well-known port) using the following
```

timwoj (changes requested):

devel/spicy/getting-started.rst
```diff
@@ -91,6 +91,21 @@ having the expected signature of four parameters matching the types of
 the parameter expressions that the ``*.evt`` file specifies. The
 handler's body then just prints out what it gets.
 
+Further, you see a ``zeek_init()`` event handler that calls the Analyzer
+framework's ``register_for_port()`` function. This call tells Zeek to
```
rsmmr, line 95: Would be nice to link to the framework and function here in Zeek's docs.

devel/spicy/getting-started.rst
```diff
@@ -91,6 +91,21 @@ having the expected signature of four parameters matching the types of
...
+enable the ``spicy::MyHTTP`` analyzer for connections on TCP port 12345.
+
+.. note::
+
+  .. versionchanged:: 7.1
```
rsmmr, line 100: I'm torn on this: we don't use `versionchanged` anywhere else but usually just update the docs and otherwise rely on NEWS. But I see how this can be helpful of course, and maybe we should start doing this. So ok to keep I guess. :-)

devel/spicy/tutorial.rst
```diff
@@ -82,34 +82,43 @@ protocol analyzer. We follow the same scheme here and put the
...
+script snippet:
+
+.. literalinclude:: examples/tftp-register-analyzer.zeek
+    :language: zeek
+
+This snippet can generally be assumed to be part of the analyzer's scripts.
+We'll reference it as ``tftp-register-analyzer.zeek`` below.
```
rsmmr, line 103:
```suggestion
We'll refer to it as ``tftp-register-analyzer.zeek`` below.
```

rsmmr (commented):
Thanks for going through!

One question: what do we do with `%port` inside a Spicy unit? That actually still works with a Zeek analyzer, too, but I don't think it's documented anywhere afaict.

devel/spicy/tutorial.rst
```diff
@@ -82,34 +82,43 @@ protocol analyzer. We follow the same scheme here and put the
...
+
 With this in place, we can already employ the analyzer inside Zeek. It
 will not generate any events yet, but we can at least see the output of
 the ``on %done { print self; }`` hook that still remains part of the
 grammar from earlier:
 
 .. code::
 
-    # zeek -r tftp_rrq.pcap tftp.hlto Spicy::enable_print=T
+    # zeek -r tftp_rrq.pcap tftp.hlto tftp-register-analyzer.zeek Spicy::enable_print=T
```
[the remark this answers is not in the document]
awelzel: > (I have to say: I liked being able to not need a Zeek script for this example ...)

I also thought that `port` / `ports` was kind of nice for the whole tutorial. Oh well :-(

Had a `zeek ... -e 'Analyzer::register_for_port(...)` in mind, but probably more confusing in the end than having it in a scirpt.

I'm tempted to name the script `tftp.zeek` like all the other examples, so it stands out a bit less (and the other examples build upon each other, so that makes some sense).

devel/spicy/getting-started.rst
awelzel: I've used `versionadded` a bit in the Zeek docs as it's nice for people that look at docs.zeek.org "master" version online to go "hmm, which version are we running, actually?" and "hmm, we should upgrade" nudge :-)

image

awelzel:
> One question: what do we do with %port inside a Spicy unit? That actually still works with a Zeek analyzer, too, but I don't think it's documented anywhere afaict.

Benjamin said that is still used for batch file processing on the Spicy side (I don't know so much about it) - so might make sense to keep it for units. On the Zeek side it would only make sense to deprecate that now, too (or even just remove it).

devel/spicy/tutorial.rst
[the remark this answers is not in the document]
rsmmr: > I'm tempted to name the script tftp.zeek like all the other examples, 

Good though, agree.

awelzel:
I'm going to close this after chatting with Robin - he'd like to re-add  `port` and `ports` support in .evt with different spin by adding a new event that's raised for loaded/installed spicy analyzers, providing metadata, including ports.

https://github.com/zeek/zeek/issues/3881

PR #275, opened 2024-09-24
Update TFTP Zeek script file for spicy tutorial

evantypanski:
I changed this file in https://github.com/zeek/spicy-tftp/pull/12 but didn't realize that it would need a docs update. The actually text of the file isn't anywhere, but it does appear in a download :)

This is causing a CI failure for Zeek in https://github.com/zeek/zeek/pull/3949


bbannier (approved):
Yep, that makes sense 😅. I just made a pass over the docs to confirm again that we do not reference particular regions of `tftp.zeek` (we do not).

PR #276, opened 2024-10-09
Fix a couple Spicy doc nits

evantypanski:
- convenient was misspelled
- Spicy TFTP example needs [another](https://github.com/zeek/zeek-docs/pull/275) pass, apparently there's another script that is roughly similar that still used `get_tag` :)
- Fixed up some whitespace stuff in that TFTP example - that file mixed tabs and 8 spaces so I just moved it to 4 spaces.

Fun fact: building docs fresh was fine, but when I pulled main and rebuilt it, it crashed my laptop and tried to crash it again when I retried. Like 7 python processes each taking 3GB :O

### DIFF
--- a/devel/spicy/examples/tftp-schedule-analyzer.zeek
+++ b/devel/spicy/examples/tftp-schedule-analyzer.zeek
@@ -1,37 +1,37 @@
 
-function schedule_tftp_analyzer(id: conn_id) 
-	{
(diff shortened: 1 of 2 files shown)


bbannier (approved):
Thanks!

The CI failures are due to e4f1f2b245fb451c30f965d4ac61a565b97c19a9 and not this PR.

PR #279, opened 2024-10-15
Fix installation of Python packages in generate docs CI job

bbannier:
This job uses `ubuntu-latest`, i.e., it will automatically use the
newest version of Ubuntu available. On current versions `pip install`
refuses to install packages into the system since this might break
system packages. Users are expected to explicitly opt into this
potential breakage by passing the `--break-system-packages` flag. This
is an acceptable workaround here since we do not distribute this setup,
and currently things appear to not break.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -18,6 +18,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Fetch Dependencies
-        run: sudo pip3 install -r requirements.txt
+        run: sudo pip3 install --break-system-packages -r requirements.txt
       - name: Generate Docs
         run: make SPHINXOPTS="-W --keep-going"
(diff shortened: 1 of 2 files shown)


bbannier:
This PR fixes `pip`-related failures. We still run into a Sphinx build error which should is addressed by https://github.com/zeek/zeek/pull/3976 and should pass here once the docs have been automatically bumped.

evantypanski (approved):
Seems better than just failing the CI job so cool

Only nit: The `requirements.txt` file locks into specific versions *except* for `pygments`, should we lock that in for the rare possibility that they break a system package?

```
Pygments>=2.12.0
```

Don't think it'd entirely eliminate the possibility, but may slightly reduce it

bbannier:
Good catch with Pygments which is indeed already present in the `ubuntu` base image. I pushed a patch which fixes is to the exact version of the image. Should we upgrade the image I'd expect this to fail since `pip install` cannot remove the system package (just happened to me when I used a different version).

PR #281, opened 2024-10-21
generate-docs: Use ubuntu-latest

awelzel:
Commit https://github.com/zeek/zeek-docs/commit/bd300424466d4939ef8e2eaa9936a09cbd59f1f5 removed the sudo usage
which was the actual culprit for using the wrong pip version. Bump up
to ubuntu-latest again.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - 'ubuntu-22.04'
+          - 'ubuntu-latest'
           - 'macos-latest'
         python-version:
           - '3.8'


awelzel:
@bbannier - see anything wrong with this? It's a follow-up to #280 based on the [comment I made there](https://github.com/zeek/zeek-docs/pull/280#discussion_r1808290459). I'll update the commit to be a bit more descriptive if you think this is good.

bbannier:
> @bbannier - see anything wrong with this? It's a follow-up to #280 based on the [comment I made there](https://github.com/zeek/zeek-docs/pull/280#discussion_r1808290459). I'll update the commit to be a bit more descriptive if you think this is good.

Not really sure how this failed before when we were already using `actions/setup-python@v5` and were als setting `python-version`; AFAICT this should never have exposed us to issue from the base image (https://github.com/actions/runner-images/issues/10636), but it did. By using an unpinned image we allow GH to break us again, and I do don't think we get much benefit from defaulting to the latest image beyond not having to worry about that image being retired.

awelzel:
> Not really sure how this failed before when we were already using actions/setup-python and were als setting python-version

Pretty sure that because of the `sudo` usage it would reset `PATH` and use the system `pip` rather than `pip` provided by `setup-python`. So I think the right fix was to remove the `sudo`.

bbannier (approved):
>> Not really sure how this failed before when we were already using actions/setup-python and were als setting python-version

> Pretty sure that because of the `sudo` usage it would reset `PATH` and use the system `pip` rather than `pip` provided by `setup-python`. So I think the right fix was to remove the `sudo`.

🤦.

Still not fully convinced of the benefit of "latest", but I guess it now might hurt less.

Do you have an idea why we need to "test" this against so many different versions? With your explanation it might not have tested any, and I also do not understand why we would care at all as long as we can build on RTD and with our minimum supported version (if at all).

PR #288, opened 2025-01-01
Fix incorrect use of `zeek:see`

bbannier:


evantypanski (approved):
Unrelated, but looking at this I also checked out uses of `zeek:see` and found the [`dns_request` event](https://github.com/zeek/zeek-docs/blob/387a01435e18d98c8ffed997d2f41fa57eda5dd6/scripting/basics.rst?plain=1#L327-L352) is missing the `original_query` parameter in the scripting basics section. May be worthwhile to update that documentation eventually

PR #294, opened 2025-01-23
Add sphinx configuration key for readthedocs

timwoj:
RTD started enforcing the existence of this key in their yml file as of 2025/01/20. See https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/ for more info. https://readthedocs.org/projects/zeek-docs/builds/26952573/ is a build that includes the config option for testing.

Our master builds are currently failing due to the lack of this key.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -11,3 +11,6 @@ build:
 python:
   install:
     - requirements: requirements.txt
+
+sphinx:
+  configuration: conf.py


awelzel (approved):
Spicy has the key already, Broker does not. Any other readthedocs projects that may need a look?

timwoj:
`zkg` is the only other one that uses readthedocs and it has it already. I'll open a PR on the broker side do a build there for sanity checking.

PR #299, opened 2025-02-27
Link Spicy in subcomponent documentation

bbannier:
It is probably worth it to directly link Spicy in the Zeek docs. In this patch I inserted a link above the link to binpac since today we would prefer users looking into Spicy instead binpac, but other than that I have no strong opinion on the exact order.


timwoj (approved):
I hadn't reviewed yesterday because it was marked as a draft. Looks fine to me.

PR #301, opened 2025-03-08
Minor fix in bifs' return types

anthonykasza:
It appears as though the descriptions for `double_to_count()` and `double_to_int()` were swapped.


timwoj:
This had to be fixed in the bif file since the docs are generated from there. See https://github.com/zeek/zeek/pull/4274.

PR #306, opened 2025-04-24
Add typos pre-commit rule, fix findings

### LINKED ISSUE #305: Add typos pre-commit
In the main Zeek repository, we have a pre-commit rule that checks for typos in comment strings and such. I'm somewhat surprised that we hadn't already added it to this repo too. It will take some work to get the set of exclusions correct so that it's not fixing things it shouldn't, but a quick test run finds a lot of worthy changes. This is the rule that needs to be added:

```
- repo: https://github.com/crate-ci/typos
  rev: v1.30.1
  hooks:
    - id: typos
```

See https://github.com/zeek/zeek/blob/master/.typos.toml for an example of the exclusions file.

timwoj:
This ended up not being as wild as I expected. It uses a modified version of the typos.toml file from zeek/zeek. A bunch of rules from there don't matter here (such as all of the script_opt `ot` ones). There are a few new rules, but nothing wild:

- Ignore `Ninteenth`, `fo(o`, `Not-ECT`, Connection and file IDs, and a couple of strings from hexdumps in some of the logs.
- Keep it from changing "Smoot" to "Smooth"

Fixes #305

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,8 @@ repos:
   - id: rst-backticks
     stages: ["pre-commit"]
     exclude: 'scripts/builtin-plugins/Zeek_JavaScript/__load__.zeek.rst'
+
+- repo: https://github.com/crate-ci/typos
+  rev: v1.30.1
+  hooks:
+    - id: typos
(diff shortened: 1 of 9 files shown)


stevesmoot:
The Smooth switcheroo is surprisingly common!

PR #307, opened 2025-04-24
Add higher-level documentation for storage framework

timwoj:


frameworks/storage.rst
```diff
@@ -0,0 +1,186 @@
...
+more examples, see the test cases in ``testing/btest/scripts/base/frameworks/storage`` and
+an example storage plugin test in ``testing/btest/plugin/storage-src``.
+
+Terminology
+===========
+
+Zeek's storage framework uses two main components:
+
+  Backend
+    A backend plugin provides access to a storage system, called a ``store`` in the rest
```
awelzel, line 22: Total nitpickery: The doc doesn't use "store" in that manner all that much (maybe twice), so maybe just ditch the note about "store"? The examples also assign the result of `open_backend()` to a variable called `backend` rather than `store`, so that just seems a bit inconsistent. (personally like backend and/or backend instance more).

frameworks/storage.rst
```diff
@@ -0,0 +1,186 @@
...
+operation has no concept of what the underlying storage architecture is. It's just using
+the handle that was opened.
+
+First, we make a call to :zeek:see:`Storage::Sync::put`, passing a key and a value to be
+stored. These must be of the same type that were passed in the arguments to
+``open_backend``, as described in the :ref:`earlier section <storage-opening-closing>`.
+The arguments passed into ``put`` are contained in a record of type
+:zeek:see:`Storage::PutArgs`. See the documentation for that type for descriptions of the
+fields available. In this case, we specify a key and a value plus an expiration time. This
+expiration time indicates when the data should be automatically removed from the store. We
```
awelzel, line 142:
```suggestion
expiration time indicates when the data should be automatically removed from the backend. We
```

Motivated by variable name and note about store vs backend earlier?

frameworks/storage.rst
```diff
@@ -0,0 +1,186 @@
...
+    return;
+    }
+
+Events
+======
+
+Two events exist for the storage framework: :zeek:see:`Storage::backend_lost` and
+:zeek:see:`Storage::backend_opened`. Both events were mentioned in the :ref:`example of
+opening and closing a backend <storage-opening-closing>`, but an additional point needs to
+be made about the :zeek:see:`Storage::backend_lost` event. This event is also posted when
```
awelzel, line 184:
```suggestion
be made about the :zeek:see:`Storage::backend_lost` event. This event is also raised when
```

awelzel (commented):
Great to have something for this already. Lots of small commentary, sorry.

frameworks/storage.rst
```diff
@@ -0,0 +1,186 @@
+.. _framework-storage:
+
+=================
+Storage Framework
+=================
+
+The storage framework provides a plugin-based system for short- and long-term storage of
+data, accessible from Zeek script-land. This is not packet data itself, but data artifacts
+generated from the packet data. It has interchangeable asynchronous and synchronous
+modes.
```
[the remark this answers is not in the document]
timwoj: My only concern about mentioning more advanced query options is that we also set the an expectation that those might exist in the future.

Added this sentence that the end of that paragraph, but without the rest:

> The framework provides just a simple key-value store, using Zeek values as the keys
> to store and lookup data.

frameworks/storage.rst
```diff
@@ -0,0 +1,188 @@
...
+Opening a backend starts with defining a set of options for that backend. The
+:zeek:see:`Storage::BackendOptions` is defined with some fields by default, but loading a
+policy for a specific backend type may add new fields to it. In the example below, we
+loaded the SQLite policy, which adds a new ``sqlite`` field with additional options. These
+options are filled in to denote where to store the sqlite database file and what table to
+use. This allows users to separate different instances of a backend from each other in a
+single database file.
+
+The script then sets a serializer. The storage framework sets this to the JSON
+(:zeek:see:`Storage::STORAGE_SERIALIZER_JSON`) serializer by default, but setting it is
```
awelzel, line 87:
```suggestion
(:zeek:see:`Storage::STORAGE_SERIALIZER_JSON`) serializer by default, but setting it explicitly is
```

awelzel (approved):

frameworks/storage.rst
```diff
@@ -0,0 +1,188 @@
...
+and options record with the :zeek:see:`Storage::Async::open_backend`/
+:zeek:see:`Storage::Sync::open_backend` functions.
+
+Operation Return Values
+-----------------------
+
+All backend methods return a record of type :zeek:see:`Storage::OperationResult`. This
+record contains a code that indicates the result of the operation. For failures, backends
+may provide more details in the optional error message. The record will also contain data
+For operations that return values, namely ``open_backend`` or ``get``.
```
ckreibich, line 67:
```suggestion
for operations that return values, namely ``open_backend`` or ``get``.
```

frameworks/storage.rst
```diff
@@ -0,0 +1,188 @@
...
+Calling :zeek:see:`Storage::Sync::open_backend` instantiates a backend connection. As
+described above, ``open_backend`` returns a :zeek:see:`Storage::OperationResult`. On
+success, it stores the handle to the backend in the ``value`` field of the result
+record. We check the ``code`` field as well to make sure the operation succeeded.  Backend
+handles can be stored in global values just like any other value. They can be opened
+during startup, such as in a :zeek:see:`zeek_init` event handler, and reused throughout
+the runtime of Zeek. When a backend is successfully opened, a
+:zeek:see:`Storage::backend_opened` event will be emitted.
+
+The two type arguments to open_backend define the script-level types for keys and
```
ckreibich, line 99:
```suggestion
The two type arguments to ``open_backend`` define the script-level types for keys and
```

frameworks/storage.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Storing, Retrieving, and Erasing Data
+-------------------------------------
+
+The true point of the storage framework is to store and retrieve data. This example shows
+making synchronous calls to add a new key/value pair to a backend, retrieve it, and erase
+the entry associated with the key. This assumes that the ``backend`` variable used below
+points to an opened backend handle. The idea is that users do not need to worry about the
+underlying backend implementation. SQLite, Redis, or other backends should behave
+identically.
```
ckreibich, line 135:
```suggestion
underlying backend implementation. In terms of Zeek's script-layer API, SQLite, Redis,
or other backends should behave identically.
```
This is a bit tricky because as an absolute statement it may not hold. For example in a multi-host cluster SQLite will conceptually behave differently from centralized Redis. That's why I think your idea of adding backend-specific sections is a great idea — you could elaborate on this there.

frameworks/storage.rst
```diff
@@ -0,0 +1,188 @@
...
+
+First, we make a call to :zeek:see:`Storage::Sync::put`, passing a key and a value to be
+stored. These must be of the same type that were passed in the arguments to
```
ckreibich, line 138:
```suggestion
stored. These must be of the same types that were passed in the arguments to
```

ckreibich (approved):
This is great Tim, nice and direct. I only have nits here.

If you're open to adding more meat in a few places, I think a bit more on the concept of expiration and more detail on individual backends (as you already suggested) would be great: centralized storage vs host-local, maybe a few words on concurrency, etc.

PR #308, opened 2025-05-02
Remove `spicy::` prefix name recommendation

evantypanski:
This seems to be stale and no longer recommended. I did not update the analyzer examples which use it, or explanation of what happens if you do use it, since those are more prevalent and reflect current use.

For example, I still think it's useful to explain what happens to `::` in the `Analyzer::Tag` enum.

Related:

https://github.com/zeek/package-template/pull/48

https://github.com/zeek/package-template/issues/47


bbannier (approved):
Thanks! 🚀

PR #309, opened 2025-05-05
Alphabetize script types section

### LINKED ISSUE #274: Types section in table of contents should be alphabetic
Other sections might need to be the same but I didn't check. This one just stood out while browsing the docs.

Screenshot 2024-09-13 at 12 50 22 PM

evantypanski:
I was looking at the script types section and got annoyed. Fixes #274 

Only other section I thought would be nice to have alphabetized was [directives](https://docs.zeek.org/en/master/script-reference/directives.html), but I like `@if` and `@else` etc. being near its related parts, so I left it.

Content changes are separated to fix some minor issues I found. From what I could tell this otherwise had no significant changes, but I would appreciate a second glance in case it turns out I don't know the alphabet :)

Screenshot 2025-05-05 at 4 54 28 PM

### DIFF
--- a/script-reference/types.rst
+++ b/script-reference/types.rst
@@ -341,7 +341,7 @@ Enumerations may assign :zeek:type:`count` values explicitly:
     type color: enum { Red = 10, White = 20, Blue = 30 };
 
 Without explicit assignment, Zeek numbers enumerations sequentially starting
-from 0. You may not mix explicit and implicit assignment.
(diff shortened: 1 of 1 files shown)


timwoj (approved):
I assume most of this is just a copy/paste job. I didn't look through all of it for any errant wording changes.

evantypanski:
> I assume most of this is just a copy/paste job.

Yep, what wasn't is in the second commit

PR #311, opened 2025-05-06
Move tracing-events into ./scripting

awelzel:
@evantypanski - what do you think about just doing that now?

I made Vern the author as I have no beef in the content :-)


scripting/index.rst
```diff
@@ -10,4 +10,5 @@ Introduction to Scripting
    usage
    event-groups
    optimization
+   tracing-events
```
evantypanski, line 13: nit: I'd personally prefer this go above optimization since it's tangentially related to event-groups

evantypanski (approved):
I prefer this by a lot!

PR #312, opened 2025-05-08
New quickstart

### LINKED ISSUE #310: Revisit quick start guide
I started looking at the quick start (and beginning of documentation) a bit and think there are a few issues:

1) The first two items in the Table of Contents are long-winded explanations about technical details ("About Zeek" and "Monitoring With Zeek"). These have valuable information, but it seems like we can "show not tell" first. If not, we can add smaller explanations. Right now they're relatively wall-of-texty. Suggestion: move "Get Started" above these sections
2) The Installing section can maybe be split up - since building from source takes up such a large portion, I would prefer that be a different section. Most users should not be building from source (is this controversial?) 
3) The quick start guide is overall long-winded and takes a while before getting to real output. There are also sections I think go better elsewhere - yes `zeekctl` may be important, but it should be explained after you see what Zeek does in the first place

I attached a little (very rough) draft for a new quick start - [quickstart-proposal.md](https://github.com/user-attachments/files/20062752/quickstart-proposal.md)

A lot of the info that isn't covered can then just be shuffled around - this issue doesn't have to encapsulate that. For example, I don't think these belong in a "quick start":

- [Tracing Events](https://docs.zeek.org/en/master/quickstart.html#tracing-events)
- [Running Zeek Without Installing](https://docs.zeek.org/en/master/quickstart.html#telling-zeek-which-scripts-to-load) - probably in the building from source section?

Other suggestions or ideas would be nice. To me, the current quick start doesn't make your first 30 minutes with Zeek easy, nor does it sell Zeek particularly well. Explaining ideas in detail can come later.

evantypanski:
Closes #310 

(starting as a draft)

This is a new quick start guide. There are a few major points I wanted to address:

1) It should not go in depth, but it should give a birds-eye view of important parts.
2) You should go from no-Zeek-knowledge to *something* in as little time as possible
3) Each part of the quick start builds on the last, rather than diving into a full configuration
4) It's all (hopefully) simple and easy English, so it's easy to follow with no ambiguity
5) Show, don't tell

I still want to explore adding sections or mentions in the quick start guide and toy with the example pcap file.

Also, the first part of the quick start would use a pcap file provided in the documentation. This is because one of the main points of Zeek is to take traffic that is a magical blob of nothing and turn it into something understandable. I'd even consider changing the name to like `sample.pcap` so that the quick start shows how you can understand unknown traffic quickly.

All to say: there's a fair bit more to do


evantypanski:
This should be ready, maybe small changes.

I'm a bit worried about the cluster section - it doesn't feel like there's the same "aha" moment that the other sections have, but it also seems useful to put in the quickstart in the spirit of "give an intro to everything useful"

For word count: the proposed quickstart has 1518, the old one has 2815. I think it also gets closer to useful information quicker, gives a better overview, and doesn't get as deep into details that should be one level deeper (like the [filesystem](https://docs.zeek.org/en/master/quickstart.html#filesystem-walkthrough)).

ckreibich (approved):
Right on, Evan!

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+     $ curl -X WEIRD http://zeek.org
+
+Zeek will analyze traffic over that network device. Therefore, the logs may
+have more than just the two entries found before. Entries should still appear in
+:file:`conn.log`, :file:`http.log`, and :file:`weird.log` for these commands.
+
+Scripting
+=========
+
+Zeek can also use its own scripting language in order to configure behavior and
```
awelzel, line 132:
```suggestion
You can also use Zeek's own scripting language in order to modify and extend its behavior:
```
If this is totally off from how you want this to sound, reject. Just seemed the user would be the one using the scripting language.

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+
+     $ zeek weird-test.zeek -r quickstart.pcap
+     HTTP request: GET / (192.168.1.8->192.0.78.212)
+     HTTP request: WEIRD / (192.168.1.8->192.0.78.212)
+
+Or on live traffic:
+
+.. code-block:: console
+
+     $ zeek -i en0
```
awelzel, line 157:
```suggestion
     $ zeek weird-test.zeek -i en0
```

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+
+The terminal running Zeek will print each command as it gets processed.
+
+More information on how to use Zeek’s scripting language can be found in the
+:doc:`../scripting/index` section.
+
+Managing Zeek
+=============
+
+Zeek comes packaged with ZeekControl (``zeekctl``) to manage more complex use
```
awelzel, line 174:
```suggestion
Zeek comes packaged with ZeekControl (``zeekctl``) to manage more complex deployments.
```

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+the device is ``en0``, that would look like:
+
+.. code-block:: console
+
+     [zeek]
+     type=standalone
+     host=localhost
+     interface=en0
+
+You can also further configure the ``local.zeek`` script found in
```
awelzel, line 193:
```suggestion
You can further configure the ``local.zeek`` script found in
```

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+``$PREFIX/share/zeek/site/local.zeek``. This script is loaded by ``zeekctl`` by
```
awelzel, line 194:
```suggestion
``$PREFIX/share/zeek/site/local.zeek``. ``zeekctl`` loads this script by default.
```

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+default. It is not overwritten by Zeek upgrades. Zeek without ``zeekctl`` may
+specify ``local.zeek`` in the command line in order to load its contents.
+
+Run ``zeekctl`` in order to get into a new prompt:
```
awelzel, line 198:
```suggestion
Run ``zeekctl`` in order to start an interactive prompt to manage your Zeek deployment:
```

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+
+     [worker-1]
+     type=worker
+     host=localhost
+     interface=en0
+
+     [worker-2]
+     type=worker
+     host=localhost
+     interface=en0
```
awelzel, line 304: Just using a single worker might be better in quickstart. There's too many caveats with multiple workers (proper, load-balancing, af_packet_fanout_id, ...). I'm even a bit worried that running through zeekctl is overwhelming, but there's also no real way around it for a real deployment.

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+
+     [zeek]
+     type=standalone
+     host=localhost
+     interface=en0
+
+You can also further configure the ``local.zeek`` script found in
+``$PREFIX/share/zeek/site/local.zeek``. This script is loaded by ``zeekctl`` by
+default. It is not overwritten by Zeek upgrades. Zeek without ``zeekctl`` may
+specify ``local.zeek`` in the command line in order to load its contents.
```
awelzel, line 196: While `local.zeek` not being added to a bare `zeek` invocation is definitely a known gotcha, unclear sneaking it in here is useful - maybe leave it out?

awelzel (commented):
Thanks for doing this. Had some thoughts, take them as you see fit :+1:

awelzel (approved):

get-started/quickstart.rst
evantypanski: you're right - I think just waiting until this point to mention it might be enough context without adding an extra explanation

get-started/quickstart.rst
evantypanski: I seem to have a fear of starting sentences with lowercase command names :)

Since your suggestion makes the sentence in active voice, though, I think it's better.

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+
+Zeek is often run on live traffic on a network, not just captured traffic.
+You can provide Zeek with a network interface to monitor traffic. Any traffic on
+that interface will be analyzed in order to create logs. For example, you may
+run Zeek on the ``en0`` network device (change ``en0`` to the device you want
+to monitor traffic on):
+
+.. code-block:: console
+
+     $ zeek -i en0
```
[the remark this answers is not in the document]
evantypanski: I'll use this as an opportunity to add a `note` - I like how it breaks up the page and it adds a nice explanation, I think.

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+=================
+Quick Start Guide
+=================
+
+Zeek is a network traffic analyzer. Zeek works on most modern Unix-based
+systems and does not require custom hardware. See :doc:`install` in order to
+install from pre-built binary packages, or :doc:`building-from-source` in order
+to build Zeek from source.
+
+Zeek requires some network traffic in order to run. For this guide, we’ll use
```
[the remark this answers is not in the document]
evantypanski: Changed it a little bit from the suggestion but no longer spreading lies :)

get-started/quickstart.rst
evantypanski: I agree, I only kept that one because it's commented out by default, so I don't need to tell people to only comment one out. That could cause frustration if they are only *kind of* following the guide.

Maybe I just remove mentions of uncommenting altogether - but then people might see them anyway. I'll grapple with this in my head for a bit to see if there's something cleaner, or maybe those wouldn't be issues?

get-started/quickstart.rst
```diff
@@ -0,0 +1,383 @@
...
+         "method": "WEIRD",
+         "host": "zeek.org",
+         "uri": "/",
+         ...
+     }
+
+The first entry is a simple ``GET`` request to zeek.org. The second entry has
+the same UID as the first entry in conn.log. The HTTP method was ``WEIRD``. For
+this case, Zeek will also generate a separate log for some unexpected behavior
+(or "weirds"), like an unknown HTTP method. You can find this in weird.log:
```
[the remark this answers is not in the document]
awelzel: > I just liked the word play.

Then you should keep it `WEIRD` :-)

evantypanski:
Current plan is to give this one more rundown Monday after a nice weekend to see if I catch any small mistakes, then merge. Then we can look at points 2-4 above^

This will also give anyone interested a nicer view than the gross `.rst` file

get-started/quickstart.rst
```diff
@@ -0,0 +1,395 @@
...
+The logs contain the ``WEIRD`` HTTP request.
+
+More information on using ZeekControl can be found in the
+`ZeekControl documentation`_. More information on setting up a cluster can be
+found in the :doc:`cluster-setup` section.
+
+Clusters
+========
+
+ZeekControl is also used to manage a cluster of Zeek processes. This cluster
```
timwoj, line 274:
```suggestion
ZeekControl is also used to manage a cluster of Zeek processes. A cluster
```

get-started/quickstart.rst
```diff
@@ -0,0 +1,395 @@
...
+contains many processes which analyze traffic together. Clusters are normally
+split among many machines. For this example, all nodes will be local.
```
timwoj, line 276: > normally split among many machines

I'm not sure this is actually true. I think more people run a single host with multiple processes than multi-host clusters.

timwoj (changes requested):

get-started/quickstart.rst
```diff
@@ -0,0 +1,395 @@
...
+
+The terminal running Zeek will print each command as it gets processed.
+
+More information on how to use Zeek’s scripting language can be found in the
+:doc:`../scripting/index` section.
+
+Managing Zeek
+=============
+
+Zeek comes packaged with ZeekControl (``zeekctl``) to manage more complex
```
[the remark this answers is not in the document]
evantypanski: I have a link at the end, the formula I was using is "show first, then link away at the end of the section" - which I prefer here so that there's an initial explanation. Not super hard to throw another link here, but I'd actually prefer not

get-started/quickstart.rst
evantypanski: I reworded it to not spread falsities :)

get-started/quickstart.rst
```diff
@@ -0,0 +1,395 @@
...
+
+Running Zeek
+============
+
+Open a terminal, navigate to a clean directory, and run zeek on the
+:file:`quickstart.pcap` file:
+
+   .. code-block:: console
+
+     zeek -r quickstart.pcap LogAscii::use_json=T
```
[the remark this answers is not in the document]
evantypanski: I added a note at the end of the section, since other sections don't actually use that. Do you like that? I don't want to add it every time, but when showing logs I *really* wanted to use JSON.

Also I don't want to take it for granted that they'd know. I think it should be explained somewhere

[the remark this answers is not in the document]
timwoj: Yep, that looks good to me.

timwoj (approved):

PR #314, opened 2025-05-15
Add some notes/caveats for the built-in storage backends

timwoj:


awelzel (approved):
I was thinking, do you want to put the following under the heading, 7.0 will be around for a bit and users looking at the docs will directly learn 7.0 won't have that yet.

```
.. versionadded:: 7.2
```

PR #315, opened 2025-05-20
Touchup install and quickstart

evantypanski:
Just a couple things I noticed. The install section should probably default to showing newer releases first IMO

Random point, I waffled on the wording for this, any other suggestions? I just realized I have to use `sudo` but never mention it:

> Commands which use a network device require root access in order to monitor traffic.

### DIFF
--- a/get-started/quickstart.rst
+++ b/get-started/quickstart.rst
@@ -124,6 +124,9 @@ want to monitor traffic on):
 
      $ zeek -i en0 -C
 
+Root access is typically required to run commands which monitor a network
+device.
+
 In another terminal, create the same two HTTP requests we saw earlier via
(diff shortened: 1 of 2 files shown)


get-started/quickstart.rst
```diff
@@ -124,6 +124,9 @@ want to monitor traffic on):
 
      $ zeek -i en0 -C
 
+Commands which use a network device require root access in order to monitor
+traffic.
```
timwoj, line 128: This wording confused me a little at first. I switched around the ordering a little, plus root access *may* be required but not always.

```suggestion
Root access is typically required to run commands which monitor a network device.
```

timwoj (approved):

PR #316, opened 2025-05-20
Fix get-started redirects

evantypanski:
@ekoyle pointed out that I broke the old quick start link with da1a0e0

This adds redirects for anything that lives in the get-started section


timwoj (approved):
I wonder if we should run the link checker as a separate GH workflow. It doesn't look like we can run it as part of the existing build steps, unfortunately. I ran it locally and there's a lot of links that either changed or are invalid. It can be a separate issue/PR from this though. The command to run it is `sphinx-build -M linkcheck . <outputdir>`.

ekoyle:
@evantypanski is there a manual step required after merging this for the changes to take effect?

It appears that `https://docs.zeek.org/en/master/quickstart.html` (first google hit for `zeek environment variables`) is still resulting in a 404.

bbannier:
> @evantypanski is there a manual step required after merging this for the changes to take effect?

Likely, and I cannot them in https://app.readthedocs.org/dashboard/zeek-docs/redirects/ (UI not ideal), so this probably still needs to be deployed.

https://github.com/zeek/zeek-docs/blob/d442ec5d58f8b6c47eacdbc866cc49c7f490ed29/redirects.yml#L7-L10

bbannier:
> I wonder if we should run the link checker as a separate GH workflow. It doesn't look like we can run it as part of the existing build steps, unfortunately. I ran it locally and there's a lot of links that either changed or are invalid. It can be a separate issue/PR from this though. The command to run it is `sphinx-build -M linkcheck . <outputdir>`.

We run this link checker for the Spicy docs. Running it in CI is tricky as remotes might be temporarily unavailable or might just ratelimit or even blackhole requests from GH (latter more typical). For that reason we have made in non-critical and I instead manually run it at intervals and around releases.

ckreibich:
The redirects all work now.

ckreibich:
Fwiw, I just ran `sphinx-build -M linkcheck` over here and it's struggling too. It hits rate-limiting from Github and I think it's getting confused about our links out to Github in the generated docs — for example it complains

```
(scripts/base/frameworks/logging/writers/ascii.zeek: line  201) broken    https://github.com/zeek/zeek/blob/master/scripts/base/frameworks/logging/writers/ascii.zeek#L100-L100 - Anchor 'L100-L100' not found
```
but that link [is fine](https://github.com/zeek/zeek/blob/master/scripts/base/frameworks/logging/writers/ascii.zeek#L100-L100). It flags thousands of these things — perhaps it's just getting tangled up in the rate limits.

For our handwritten docs the findings look great though, so perhaps we could use a way to just run it on those?

PR #319, opened 2025-05-22
Fix redirects (again)

evantypanski:
It turns out #316 was probably worthwhile - but we have too many redirects! It's apparently capped at 100 now, but we have over 600.

So I just moved the quickstart and whatnot back out to the top level directory. There's one redirect change for the `cross-compiling` section. That one will have to be manually found and changed, I think. I manually changed it.


evantypanski:
@bbannier does this seem proper, after looking at redirects too?

cluster-setup.rst
bbannier, line 1: You didn't undo the link to `cluster/pf_ring` at the bottom and it now seems to point outside the docs.

```patch
diff --git a/cluster-setup.rst b/cluster-setup.rst
index 14a6140..bdf734f 100644
--- a/cluster-setup.rst
+++ b/cluster-setup.rst
@@ -503,7 +503,7 @@ Head over to :ref:`cluster-pf-ring` for more details.
 .. toctree::
    :hidden:

-   ../cluster/pf_ring
+   cluster/pf_ring


 .. [#] Some Linux kernel versions between 3.10 and 4.7 might exhibit
```

bbannier (approved):
LGTM.

redirects.yml
bbannier, line 1: Would you mind adding a comment near the top that this file must not be edited since we cannot upload changes?

If you feel like adding additional context, it should be possible to _remove_ redirects, but only through the UI. If we do that we should also make a corresponding change here.

bbannier (approved):
Thanks for the fixup commit, much easier to review 😅 

🚀

PR #320, opened 2025-05-22
Address new quickstart feedback

evantypanski:
DRAFT as more feedback is gathered :)

Main things I wanted addressed:

1) Link to try.zeek.org (multiple times, in fact)
2) Use only one worker in the cluster because two is confusing

Happy to add more if any more feedback comes up. Just doing it in a draft so that PRs don't get spammed if there's more feedback

I also want to get rid of the `LogAscii::use_json=T` - but I want to get rid of it by making that default in Zeek ;)

### DIFF
--- a/quickstart.rst
+++ b/quickstart.rst
@@ -16,7 +16,8 @@ install from pre-built binary packages, or :doc:`building-from-source` in order
 to build Zeek from source.
 
 We will first analyze previously captured network traffic from a ``pcap`` file -
-:download:`quickstart.pcap <traces/quickstart.pcap>`. Later, we will use Zeek to monitor live traffic.
+:download:`quickstart.pcap <traces/quickstart.pcap>`. Later, we will use Zeek to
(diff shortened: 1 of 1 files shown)


about.rst
```diff
@@ -1,3 +1,5 @@
+:orphan:
```
ckreibich, line 1: What does that do? :-)

evantypanski: since it's not linked anywhere Sphinx got mad and warned, this silences that warning. I want the page so previous links to it continue to work, since we can't add more redirects

evantypanski:
> Sorry to rain on your parade — what do you think?

I fully agree that it's a lot for the very first thing - I was intending to have a rewrite of that section in the future, but I'm not currently willing to do it (well I don't feel like I have the historical knowledge or time at the moment)

I'll just drop the commit though, I agree

ckreibich (approved):
Thanks a bunch. :+1:

PR #321, opened 2025-05-22
Add note about requirement for hiredis to be installed

timwoj:
@initconf mentioned that the Redis backend wasn't building on his FreeBSD 12 machine. Fbsd 12 only provides hiredis 1.0.2, which doesn't include the poll adapter. That adapter isn't available until hiredis 1.1.0.

### DIFF
--- a/frameworks/storage.rst
+++ b/frameworks/storage.rst
@@ -195,6 +195,9 @@ Notes for Built-in Backends
 Redis
 -----
 
+- The Redis backend requires the ``hiredis`` library to installed on the system in order
+  to build. At least version 1.1.0 (Released Nov 2022) is required.
+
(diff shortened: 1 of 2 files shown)


awelzel (approved):

frameworks/storage.rst
```diff
@@ -195,6 +195,9 @@ Notes for Built-in Backends
 Redis
 -----
 
+- The Redis backend requires the ``hiredis`` library to installed on the system in order
+  to build. At least version 1.1.0 (Released Nov 2022) is required.
+
```
awelzel, line 200: Maybe also add the hiredis as optional dependency to building-from-source.

awelzel (approved):

PR #322, opened 2025-05-27
[RFC] ext: Add zeek:recfield role

awelzel:
This role currently accepts either a plain string, or a `Type$field`
string. For the latter, it'll be rendered as just field, but a ref
is created to `Type`. We do not have anchors to individual fields,
but if we were to improve zeekygen to produce such anchors, we could
have proper links to the fields definition instead.

This PR updates the notice framework text as a demo.

Thoughts or other ideas? Different name?

---

This might seem a bit cumbersome to type, but it should help to make field references look more consistent and potentially in the future have hyperlinks directly into the place where the field is defined. Zeekygen's output doesn't currently allow for that.


awelzel:
Replaced by #324

PR #324, opened 2025-05-28
ext/zeek: Add zeek:field role and directive

awelzel:
For documenting and referencing record fields, introduce a new role and
directive. Zeek is changed to produce ReST containing `.. field:`
directives instead of the definition-list style output.

Record fields can be referenced in rst as :zeek:field:`Type$field`. This
will produce a clickable link rendered as just `field`. The link goes to
the definition of the field within the named record type. It is also possible
to use :zeek:see:`Type$field`. In that case, the link will be `Type$field`
instead of just `field`.
Using :zeek:field:`field` without a Type is allowed as markup, but when
no link is wanted.

---

I hope this was worth it :-)

This changes the way records are rendered in the HTML docs slightly. Each field can now be linked individually, so I think that's a plus. There's also a bit more internal structure on the sphinx side rather than just treating fields as rst definitions in case we want to do more fancy things in the future.

Screenshot from 2025-05-28 15-52-31

The main difference is that the gray bar does not extend to the right anymore, but also that fields are now actually next to a `Fields` entry.

The record's docstring comes at the end after all fields, but that's already an issue with master today and I'm out of steam staring at sphinx/docutils:

Screenshot from 2025-05-28 15-54-38

Not also how enums are rendered similarly today (just without a nice "Fields").

### DIFF
--- a/script-reference/autogenerated-file-analyzer-index.rst
+++ b/script-reference/autogenerated-file-analyzer-index.rst
@@ -284,47 +284,76 @@ Types
 
    :Type: :zeek:type:`record`
 
-      version: :zeek:type:`count` :zeek:attr:`&log`
-         Version number.
 
-      serial: :zeek:type:`string` :zeek:attr:`&log`
-         Serial number.
+   .. zeek:field:: version :zeek:type:`count` :zeek:attr:`&log`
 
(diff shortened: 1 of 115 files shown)


timwoj (approved):
Looks fine to me without diving way into how sphinx works. I like it 👍🏼

PR #325, opened 2025-05-28
Add exat/EXAT from the Redis analyzer to .typos.toml

timwoj:


awelzel (approved):
Ah! That's new. I was wondering how I ran into that with #324  :-)

PR #327, opened 2025-06-04
devel: Add event metadata plugin example

awelzel:
I wonder if we should actually publish this kind of plugin somewhere officially, or integrate the functionality into Zeek as an optional feature.

I'm pretty impressed how "easy" it is to do cluster event latency monitoring using a Prometheus histogram with an external plugin now that we have all that infrastructure...


devel/plugins/event-metadata-plugin.rst
```diff
@@ -0,0 +1,101 @@
...
+
+=====================
+Event Metadata Plugin
+=====================
+
+.. versionadded:: 8.0
+
+
+Zeek's plugin API allows adding metadata to Zeek events which are then
+transported across a Zeek cluster for consumption by other Zeek nodes.
```
ckreibich, line 11: Doesn't it also allow adding metadata for anything locally published, regardless of where the event ends up?

devel/plugins/event-metadata-plugin.rst
```diff
@@ -0,0 +1,101 @@
...
+This document describes this functionality in form of a tutorial. We'll
```
ckreibich, line 12: Forgive my nittiness today! :-) But — I think we should stop using the term "document" in our documentation. It's not a loose collection of documents, like say a Google Drive. It's a section or a chapter.

ckreibich (approved):
I like it a lot. Cool example of things newly possible in Zeek, and integrating them into a modern environment. I just have a few nits — feel free to handle as you see fit.

Going forward I don't think we should add more full-blown plugins in the docs though, for a few reasons:
- We won't catch errors in them as easily as if they were packages with CI etc
- People need to clone zeek-docs to get to the source.
- If people build on this, I'd argue they are more likely to ship the result in a package.

I think it'd be very nice to have "tutorial packages" for typical package/plugin scenarios that people can study. Sort of a more fleshed-out starting point than the raw package templates, that we can then link to from the docs. @bbannier also flagged the other day that it'd be nice to focus on building out more of a "utility ecosystem" for Zeek, so this could be part of that.

If you want to keep it here for now, I'm a-okay with that. But it seems fairly straightforward to make them packages, too.

devel/plugins/event-metadata-plugin.rst
awelzel: Thanks Christian! This is actually appreciated :-)

PR #332, opened 2025-07-03
Update conn-key plugin tutorial for ctx

awelzel:


awelzel:
@rsmmr - mind looking this one over. The tutorial update is a bit noisy, but I'm mostly wondering about your thoughts about the "Use of conn_id_ctx" section - does that seem a reasonable summary as a forward-looking pattern that still needs some figuring out in practice?

scripting/conn-id-ctx.rst
```diff
@@ -0,0 +1,103 @@
...
+For example, host 10.0.0.37 in segment A and host 10.0.0.37 in segment B
+share the same IP address, but represent different systems.
+The terminology used within Zeek is that IP addresses or connections
+are observed in different "contexts".
+
+
+Commonly, ethernet layers may contain information that makes it possible to discriminate
+between different network segments. Concretely, VLAN tagging is one such approach.
+Segments can also be distinguished by Virtual Network Identifiers (VNIs)
+in case of UDP-based tunnels like VXLAN or Geneve.
```
rsmmr, line 27: Suggest to turn this around a bit, with more of the example early on. Like this:


```suggestion
In some deployments, Zeek may receive network traffic from different network segments sharing overlapping IP ranges. In such cases, there is then usually some additional means of separating out the two ranges, such as VLAN numbers. For example, host 10.0.0.37 in VLAN 1 and host 10.0.0.37 in VLAN 2 may share the same IP address, but represent different systems. In Zeek's terminology, such IP addresses (or connections) occur in different *contexts*. In the example, the context is the VLAN ID; in other settings, the context could be, say, Virtual Network Identifiers as used with UDP-based tunnels like VXLAN or Geneve. Generally, from Zeek's perspective, the context can be any kind of value that Zeek can derive from the Ethernet layers.
```

scripting/conn-id-ctx.rst
```diff
@@ -0,0 +1,103 @@
...
+
+
+Since Zeek 8.0, this information can be extracted by
+:ref:`plugin-provided connection key implementations <connkey-plugin>`
+and included into Zeek's core connection tracking. Further, plugins
+:zeek:keyword:`redefine <redef>` :zeek:see:`conn_id_ctx` with additional
+fields to expose the extracted information used for connection tracking to
+the Zeek scripting layer.
```
rsmmr, line 35:
```suggestion
Since version 8.0, Zeek can extract these contexts through
:ref:`plugin-provided connection key implementations <connkey-plugin>`
and include them into its core connection tracking. Such plugins then also
:zeek:keyword:`redefine <redef>` :zeek:see:`conn_id_ctx` with additional
fields to expose the context values to the Zeek scripting layer.
```

scripting/conn-id-ctx.rst
```diff
@@ -0,0 +1,103 @@
...
+	global connection_counts: table[conn_id_ctx, addr] of count &default=0;
+
+	event new_connection(c: connection) {
+		++connection_counts[c$id$ctx, c$id$orig_h];
+	}
+
+
+If :zeek:field:`ctx` is populated with fields for VLAN tags, the table indexing
+will take them into account and create individual entries per ``(ctx, addr)``
+pair.
```
rsmmr, line 57:
```suggestion
If, for example, :zeek:field:`ctx` is populated with fields for VLAN tags, that table will create individual entries per ``(VLAN, addr)`` pair. This will also work correctly if no context has been defined: `c$id$ctx` will then just be an empty records with no fields. (This remains the default for most Zeek installations.)
```

scripting/conn-id-ctx.rst
```diff
@@ -0,0 +1,103 @@
...
+
+
+Alternatively, users can define their own record type that includes both :zeek:see:`conn_id_ctx` and :zeek:type:`addr`,
+and use instances of such records to index into tables.
```
rsmmr, line 61:
```suggestion
and use instances of such records to index into tables:
```

scripting/conn-id-ctx.rst
```diff
@@ -0,0 +1,103 @@
...
+    [ctx=[vlan=42, inner_vlan=<uninitialized>], a=141.142.228.5], HTTP
+    [ctx=[vlan=10, inner_vlan=20], a=141.142.228.5], HTTP
+    [ctx=[vlan=<uninitialized>, inner_vlan=<uninitialized>], a=141.142.228.5], HTTP
+
+
+Note that this script snippet isn't VLAN-specific, yet it is VLAN-aware. When
+using a different connection key plugin, like the one discussed in the
+:ref:`connection key tutorial <connkey-plugin>`, the result is as follows instead,
+discriminating entries in the ``talks_with_service`` table by the value of
+``c$id$ctx$vxlan_vni``.
```
rsmmr, line 85:
```suggestion
``c$id$ctx$vxlan_vni``:
```

rsmmr (commented):
I looked (only) at the new section and suggested some edits. Feel free to polish further.

ckreibich (approved):
Thanks Arne. LGTM modulo the redundant content in `conn-id-ctx.rst`.

Don't worry too much about the details here — I'd like to make a light edit pass over these docs once you are done anyway.

awelzel:
> Don't worry too much about the details here — I'd like to make a light edit pass over these docs once you are done anyway.

Thanks. Your edit pass may as well be heavy. Hope it serves as good starting point!

PR #333, opened 2025-07-04
Spicy doc updates for #4619.

rsmmr:


devel/spicy/autogen/zeek-functions.spicy
```diff
@@ -59,6 +59,17 @@ Instructs Zeek to flip the directionality of the current connection.
 
 Returns the number of packets seen so far on the current side of the current connection.
 
+.. _spicy_has_analyzer:
+
+.. rubric:: ``function zeek::has_analyzer(analyzer: string, if_enabled: bool = True) : AnalyzerType``
```
awelzel, line 64: I'm sorry for being annoying, but `has_` returning something else than True / False is eye-brow raising API :-)

Maybe `zeek::analyzer_type()` and define Undef as not available?

```
if ( zeek::analyzer_type("SSL") != Undef )
    data_in(SSL, ...)
```

rsmmr: The idea was that you can test the return value for true/false, because `Undef` coerces to false:

```
if ( zeek::has_analyzer_type("SSL") )
    data_in(SSL, ...)
```

Getting the type of analyzer just happens to be a potentially useful capability as well. From that perspective, I'm reluctant to rename to `analyzer_type()`. I think I would rather remove the analyzer type altogether, and switch to a boolean, if that what it takes to make `has_analyzer()` acceptable. :)

rsmmr: Ok, convinced. I'll split this into two functions, `has_analyzer()` and `analyzer_type()`.

PR #334, opened 2025-07-09
log-files: Add logs added during the past years

awelzel:


awelzel:
Maybe a logschema exporter for the zeek-docs would be a nice project :-)

PR #341, opened 2025-08-06
Update for dpd.log->analyzer.log change

0xxon:


timwoj (changes requested):
Should this section go away or be changed as well? https://github.com/zeek/zeek-docs/blob/master/script-reference/log-files.rst?plain=1#L294-L296

I ask because it's part of what's breaking the docs build on the deprecation-removal branch:

```
WARNING: script-reference/log-files: unknown target for ":zeek:type:`DPD::Info`"
WARNING: script-reference/log-files: unknown target for ":zeek:type:`DPD::Info`"
```

ckreibich:
I just opened #343 for the removal, so we can just focus on adding the one needed redirect here.

PR #342, opened 2025-08-07
Fix typo in cluster config

evantypanski:
`af_packet_found_id` -> `af_packet_fanout_id` (as used in the config file directly after)


awelzel (approved):
:man_facepalming:

PR #344, opened 2025-08-11
Add WebSocket API docs to devel section

awelzel:
Phew, I'm actually not sure what there really is to document. I think I'm just too much in the weeds.

@simeonmiteff / @bbannier  - any concerns having links to the libraries in Zeek docs?


devel/websocket-api.rst
```diff
@@ -0,0 +1,152 @@
...
+connections. This allows these clients to interact with Zeek's publish-subscribe
+layer and exchange Zeek events with other Zeek nodes.
+Initially, this implementation resided in the Broker subsystem.
+With Zeek 8.0, most of the implementation has been moved into core Zeek
+itself with the v1 serialization format remaining in Broker.
+
+WebSocket clients may subscribe to a fixed set of topics and will receive
+Zeek events matching these topics that Zeek cluster nodes, but also other
+WebSocket clients publish.
+Clients
```
bbannier, line 46:
```suggestion
```

devel/websocket-api.rst
```diff
@@ -0,0 +1,152 @@
...
+   {"type":"ack","endpoint":"d955d990-ad8a-5ed4-8bc5-bee252d4a2e6-websocket-0","version":"8.0.0-dev.684"}
+   {"type":"data-message","topic":"zeek.test","@data-type":"vector","data":[{"@data-type":"count","data":1},{"@data-type":"count","data":1},{"@data-type":"vector","data":[{"@data-type":"string","data":"hello"},{"@data-type":"vector","data":[{"@data-type":"count","data":3}]},{"@data-type":"vector","data":[]}]}]}
+
+The received messages, again, are encoded in Broker's JSON format. Above ``data-message``
+represents an event received on topic ``zeek.test``. The event's name is ``hello``.
+This event has a single argument of type :zeek:type:`count`. In the example above
+its value is ``3``.
+
+To send events, WebSocket clients similarly encode their event representation
+to Broker's JSON format and send them as `text data frames <https://datatracker.ietf.org/doc/html/rfc6455#section-5.6>`_.
```
bbannier, line 125: Probably worth adding:


```suggestion
To send events, WebSocket clients similarly encode their event representation
to Broker's JSON format and send them as `text data frames <https://datatracker.ietf.org/doc/html/rfc6455#section-5.6>`_. Clients can only publish to topics they are subscribed to.
```

devel/websocket-api.rst
```diff
@@ -0,0 +1,152 @@
...
+
+
+Language Bindings
+-----------------
+
```
bbannier, line 130: Maybe worth calling out that one can interact with the Zeek websocket API using any websocket client library, and anything listed here just adds convenience features on top (e.g., serialization/deserialization; state machine management).

awelzel: Yeah, good point! I've added something in that regard.

devel/websocket-api.rst
bbannier: I think I started believing that when writing my client lib, but I do not see it anymore.

PR #348, opened 2025-08-13
Add ZeroMQ installation details to the building-from-source instructions

ckreibich:
As of Zeek 8, ZeroMQ is no longer optional. @awelzel, @timwoj, does this look right to you?

This should go into the final Zeek 8 docs version.

### DIFF
--- a/building-from-source.rst
+++ b/building-from-source.rst
@@ -45,6 +45,7 @@ development headers for libraries:
     * OpenSSL (https://www.openssl.org)
     * Python 3.9 or greater (https://www.python.org/)
     * SWIG (https://www.swig.org)
+    * ZeroMQ (https://zeromq.org)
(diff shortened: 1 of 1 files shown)


building-from-source.rst
```diff
@@ -168,14 +172,14 @@ To install these, you can use:
   unzip it, and then pass ``-DPCAP_ROOT_DIR="<path to npcap sdk>"`` to the
   initial CMake invocation for Zeek.
 
+  Note also that the ZeroMQ cluster backend is not yet supported on Windows.
```
awelzel, line 175: Think that's mostly because there's no libzmq available with `choco`. Compiling libzmq on Windows by hand would probably work.

awelzel (approved):

